PR #2, opened 2022-04-18
[MER-668] Spec out basic commands

travigd:
For consistency, the implementation is mostly:
* top level commands are defined as files (e.g., `stack.go` defines `stackCmd`)
* subcommands are defined within those files (e.g., `av stack branch` defines `stackBranchCmd`)
* flag are grouped into one-off structs (`stackBranchCmd` has a matching `stackBranchFlags`) whose names match the flags (e.g., `NoPush` becomes `--no-push`)


cmd/av/stack.go
```diff
@@ -0,0 +1,173 @@
...
+	Short: "managed stacked pull requests",
+}
+
+var stackBranchFlags struct {
+	// The parent branch to base the new branch off.
+	// By default, this is the current branch.
+	Parent string
+}
+var stackBranchCmd = &cobra.Command{
+	Use:   "branch <branch name>",
```
jainankit, line 20: According to the doc this was supposed to be:
```
av stack create bookstore-backend
```

Not av branch?

cmd/av/stack.go
```diff
@@ -0,0 +1,173 @@
...
+			return errors.New("branch name is required")
+		}
+		return errors.New("unimplemented")
+	},
+}
+
+var stackSyncFlags struct {
+	// Set the parent of the current branch to this branch.
+	// This effectively re-roots the stack on a new parent.
+	Parent string
```
jainankit, line 33: necessary?

cmd/av/stack.go
```diff
@@ -0,0 +1,173 @@
...
+	// Only valid if synchronizing the root of a stack.
+	// This effectively re-roots the stack on the latest commit from the trunk.
+	Trunk bool
+	// If set, do not push to GitHub.
+	NoPush bool
+	// If set, we're continuing a previous sync.
+	// TODO:
+	// 	 we might not actually need this, we can probably detect that
+	//   a sync needs to be completed automagically and do the right thing.
+	Continue bool
```
jainankit, line 44: yeh I don't follow how this would be used. Are you imaging this to be a wrapper on top of rebase?

cmd/av/stack.go
```diff
@@ -0,0 +1,173 @@
...
+}
+var stackSyncCmd = &cobra.Command{
+	Use:   "sync",
+	Short: "synchronize all stacked branches after the current branch",
```
jainankit, line 48: after or before? So if we do:
X -> Y -> Z
and we are on Y. Do we update Y with X or do we update Z from Y?

jainankit: I would imagine that you go a specific branch and sync everything up to this branch?

I guess the other alternative is that we push a commit to this branch and push it upwards (to the next PRs). 🤷

cmd/av/stack.go
```diff
@@ -0,0 +1,173 @@
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return errors.New("unimplemented")
+	},
+}
+
+var stackTreeFlags struct {
+	// Print the stack starting at this branch.
+	// If not set, we start at the base branch of the repository.
+	Root string
```
jainankit, line 57: seems unncessary. It's okay to define these things for future, but I really don't think we need to provide that many custom options in v1. We can just print the full stack.

cmd/av/stack.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		return errors.New("unimplemented")
+	},
+}
+
+var stackStackStatus = &cobra.Command{
+	Use:   "status",
+	Short: "show the status of the stack",
+	Long:  `Show the status of the stack.`,
```
jainankit, line 120: why have a long desc? I don't see elsewhere.

cmd/av/stack.go
```diff
@@ -0,0 +1,173 @@
...
+		if n <= 0 {
+			return errors.New("invalid number (must be >= 1)")
+		}
+
+		return errors.New("unimplemented")
+	},
+}
+
+var stackStackStatus = &cobra.Command{
+	Use:   "status",
```
jainankit, line 118: seems unnecessary. We should just show this with `tree` itself.

cmd/av/stack.go
travigd: Aye, I thought `branch` would be more intuitive. `stack create` makes it sound like you need to run the command before starting the stack, whereas `stack branch` is more intuitive I think (you run the command when you want to branch off the current thing you're working on).

cmd/av/stack.go
travigd: Yeah. This is how you'd be able to tell Aviator that you want to change your branch to be based on something else. So if you do some work then realize you need to actually make it into a stacked PR, you can commit and run `av stack sync --parent my-feature`.

cmd/av/stack.go
travigd: Yeah. That's the whole point of the `stack sync` command. You'd do some work, commit, and then run `stack sync` to propagate your changes.

cmd/av/stack.go
travigd: It updates every branch *after* the current branch. I think that makes sense considering that you'll generally commit to a branch then update all the children.

I suspect you're thinking of a workflow where you edit `X`, then want to edit `Y`, then want to edit `Z`. With this it would be kind of difficult (you'd have to edit X, then run sync -- which rebases everything downstack, then checkout and edit Y, run sync, etc., which means you might be resolving a lot of unnecessary conflicts).

Yeah, let me think through this a bit more.

cmd/av/stack.go
travigd: I do want a way to view just the status of the current branch within the stack (think that will be useful) and means we can keep the `tree` view relatively minimal.

cmd/av/stack.go
travigd: Whoops, will fix.

FYI, the short is what shows up in the `Available Commands` section:
```
managed stacked pull requests

Usage:
  av stack [command]

Available Commands:
  branch      create a new stacked branch
  next        checkout the next branch in the stack
  prev        checkout the previous branch in the stack
  status      show the status of the stack
  sync        synchronize all stacked branches after the current branch
  tree        show the tree of stacked branches

Flags:
  -h, --help   help for stack

Global Flags:
      --debug   enable verbose debug logging

Use "av stack [command] --help" for more information about a command.
```

The long is what shows up in the help text.
```
go run ./cmd/av stack status --help
Show the status of the stack.

Usage:
  av stack status [flags]

Flags:
  -h, --help   help for status

Global Flags:
      --debug   enable verbose debug logging
```

cmd/av/stack.go
jainankit: That's fair. I also like `av branch`, feels similar to `git branch`.

cmd/av/stack.go
jainankit: Makes sense. Perhaps then the comment should be more describing this as adding the branch to the stack.
A secondary behavior is that this can be also used to move the branch somewhere else in stack, but that's the impression I got reading the comment.

cmd/av/stack.go
jainankit: I think we already discussed this offline.

cmd/av/stack.go
jainankit: I'd prefer if we keep both these params out for now.

jainankit (approved):

PR #3, opened 2022-04-20
[MER-671] Implement basic av stack branch command

travigd:
```
$ go run ./cmd/av -C ~/Aviator/mergequeue-demo/ stack branch foobar      
Created branch foobar

$ go run ./cmd/av -C ~/Aviator/mergequeue-demo/ stack branch foobar      
error: cannot create branch foobar (does it already exist?)
```

### DIFF
--- a/cmd/av/main.go
+++ b/cmd/av/main.go
@@ -1,12 +1,14 @@
 package main
 
 import (
+	"emperror.dev/errors"
 	"fmt"
 	"github.com/aviator-co/av/internal/config"
 	"github.com/aviator-co/av/internal/git"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"os"
+	"os/exec"
 	"strings"
 )
 
@@ -70,18 +72,19 @@ func main() {
 	}
 }
 
-func indent(s string, indent string) string {
+func indent(s string, prefix string) string {
 	// why is this not in the stdlib????
-	return "  " + strings.Replace(s, "\n", "\n"+indent, -1)
+	return prefix + strings.Replace(s, "\n", "\n"+prefix, -1)
 }
 
 func getRepo() (*git.Repo, error) {
+	cmd := exec.Command("git", "rev-parse", "--show-toplevel")
 	if rootFlags.Directory != "" {
-		return git.OpenRepo(rootFlags.Directory)
+		cmd.Dir = rootFlags.Directory
 	}
-	wd, err := os.Getwd()
+	toplevel, err := cmd.Output()
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "failed to determine repo toplevel")
 	}
-	return git.OpenRepo(wd)
+	return git.OpenRepo(strings.TrimSpace(string(toplevel)))
 }
--- a/cmd/av/main.go
+++ b/cmd/av/main.go
@@ -3,13 +3,16 @@ package main
 import (
 	"fmt"
 	"github.com/aviator-co/av/internal/config"
+	"github.com/aviator-co/av/internal/git"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"os"
+	"strings"
 )
 
 var rootFlags struct {
-	Debug bool
+	Debug     bool
+	Directory string
 }
 
 var rootCmd = &cobra.Command{
@@ -40,6 +43,10 @@ func init() {
 		&rootFlags.Debug, "debug", false,
 		"enable verbose debug logging",
 	)
+	rootCmd.PersistentFlags().StringVarP(
+		&rootFlags.Directory, "", "C", "",
+		"directory to use for git repository",
+	)
 	rootCmd.AddCommand(
 		prCmd,
 		stackCmd,
@@ -49,15 +56,32 @@ func init() {
 
 func main() {
 	if err := rootCmd.Execute(); err != nil {
-		format := "error: %s\n"
 
 		// In debug mode, show more detailed information about the error
 		// (including the stack trace if using pkg/errors).
 		if rootFlags.Debug {
-			format = "error: %#+v\n"
+			stackTrace := fmt.Sprintf("%+v", err)
+			_, _ = fmt.Fprintf(os.Stderr, "error: %s\n%s\n", err, indent(stackTrace, "\t"))
+		} else {
+			_, _ = fmt.Fprintf(os.Stderr, "error: %s\n", err)
 		}
 
-		_, _ = fmt.Fprintf(os.Stderr, format, err)
 		os.Exit(1)
 	}
+}
+
+func indent(s string, indent string) string {
+	// why is this not in the stdlib????
+	return "  " + strings.Replace(s, "\n", "\n"+indent, -1)
+}
+
+func getRepo() (*git.Repo, error) {
+	if rootFlags.Directory != "" {
+		return git.OpenRepo(rootFlags.Directory)
+	}
+	wd, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
+	return git.OpenRepo(wd)
 }
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -85,23 +85,34 @@ func (r *Repo) CurrentBranchName() (string, error) {
 	return r.Git("symbolic-ref", "--short", "HEAD")
 }
 
-// CheckoutWithCleanup performs a checkout of the given branch and returns a function that can be
-// used to revert the checkout. This is useful when switching to another branch and switching back
-// during cleanup.
-func (r *Repo) CheckoutWithCleanup(args ...string) (func(), error) {
-	currentBranch, err := r.CurrentBranchName()
+type CheckoutBranch struct {
+	// The name of the branch to checkout.
+	Name string
+	// Specifies the "-b" flag to git.
+	// The checkout will fail if the branch already exists.
+	NewBranch bool
+}
+
+// CheckoutBranch performs a checkout of the given branch and returns the name
+// of the previous branch, if any (this can be used to restore the previous
+// branch if necessary). The returned previous branch name may be empty if the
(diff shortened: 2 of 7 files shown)


cmd/av/main.go
```diff
@@ -40,6 +43,10 @@ func init() {
 		&rootFlags.Debug, "debug", false,
 		"enable verbose debug logging",
 	)
+	rootCmd.PersistentFlags().StringVarP(
```
jainankit, line 46: From the reference it seems like you have to provide:
Directory, "argument_name", "short_code", "", "description"). E.g. from Cobra:
https://cobra.dev/

Am I looking at the wrong thing?
```
StringVarP(&Source, "source", "s", "", "Source directory to read from")
```

jainankit: Also seems like an odd thing to ask for, this should be just part of some config file?

cmd/av/main.go
```diff
@@ -49,15 +56,32 @@ func init() {
 
 func main() {
 	if err := rootCmd.Execute(); err != nil {
-		format := "error: %s\n"
 
 		// In debug mode, show more detailed information about the error
 		// (including the stack trace if using pkg/errors).
 		if rootFlags.Debug {
-			format = "error: %#+v\n"
+			stackTrace := fmt.Sprintf("%+v", err)
```
jainankit, line 63: there was a `#` that's removed from the formatting, intentional?

cmd/av/main.go
```diff
@@ -49,15 +56,32 @@ func init() {
...
 		}
 
-		_, _ = fmt.Fprintf(os.Stderr, format, err)
 		os.Exit(1)
 	}
 }
+
+func indent(s string, indent string) string {
+	// why is this not in the stdlib????
+	return "  " + strings.Replace(s, "\n", "\n"+indent, -1)
```
jainankit, line 75: nit: odd to define method name and variable name same.

cmd/av/main.go
```diff
@@ -49,15 +56,32 @@ func init() {
...
+}
+
+func getRepo() (*git.Repo, error) {
+	if rootFlags.Directory != "" {
+		return git.OpenRepo(rootFlags.Directory)
+	}
+	wd, err := os.Getwd()
```
jainankit, line 82: there must be a better way in git-world to do this:
```
git rev-parse --show-toplevel
```

internal/utils/cleanup/cleanup.go
```diff
@@ -0,0 +1,20 @@
+package cleanup
+
+// Cleanup provides an easy way to clean up resources after an operation fails.
+type Cleanup struct {
+	fns []func()
+}
+
+func (c *Cleanup) Add(fn func()) {
+	c.fns = append(c.fns, fn)
```
jainankit, line 9: what are we cleaning here?

internal/git/git.go
```diff
@@ -0,0 +1,130 @@
...
+	return r, nil
+}
+
+func (r *Repo) DefaultBranch() string {
+	return r.defaultBranch
+}
+
+func (r *Repo) Git(args ...string) (string, error) {
+	cmd := exec.Command("git", args...)
+	cmd.Dir = r.repoDir
```
jainankit, line 45: why do you need to specify a directory? Assuming someone is running this from the git repo, does it matter where in repo they are?

internal/git/git.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// CurrentBranchName returns the name of the current branch.
+// The name is return in "short" format -- i.e., without the "refs/heads/" prefix.
+func (r *Repo) CurrentBranchName() (string, error) {
+	return r.Git("symbolic-ref", "--short", "HEAD")
+}
+
+// CheckoutWithCleanup performs a checkout of the given branch and returns a function that can be
+// used to revert the checkout. This is useful when switching to another branch and switching back
```
jainankit, line 89: Isn't it simpler design to just keep the branch name in memory with the caller?

internal/git/utils.go
```diff
@@ -0,0 +1,8 @@
+package git
+
+func ShortSha(sha string) string {
+	if len(sha) > 7 {
```
jainankit, line 4: sha should always be > 7.

internal/stacks/stacks.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, errors.WrapIff(
+			err,
+			"failed to create squash commit for branch",
+		)
+	}
+	logrus.WithField("sha", git.ShortSha(squashCommit)).Debug("created synthetic squash commit")
+
+	// Save the squash commit as a new branch.
+	baseRefName := "refs/av/stack-base/" + opts.Name
+	if err := repo.UpdateRef(&git.UpdateRef{Ref: baseRefName, New: squashCommit}); err != nil {
```
jainankit, line 91: how come we don't care about Old ID here but only in the head branch?

internal/stacks/stacks.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	if err := writeStackMetadata(repo, next); err != nil {
+		return nil, err
+	}
+	cu.Cancel()
+	return next, nil
+}
+
+func writeStackMetadata(repo *git.Repo, s *Branch) error {
+	refName := "refs/av/stack-metadata/" + s.Name
```
jainankit, line 125: where would we use this?

internal/stacks/stacks.go
```diff
@@ -0,0 +1,141 @@
...
+	// Save the squash commit as a new branch.
+	baseRefName := "refs/av/stack-base/" + opts.Name
+	if err := repo.UpdateRef(&git.UpdateRef{Ref: baseRefName, New: squashCommit}); err != nil {
+		return nil, err
+	}
+	logrus.WithField("ref", baseRefName).Debug("created base ref")
+
+	// Create the new branch and check it out
+	headRefName := "refs/heads/" + opts.Name
+	if err := repo.UpdateRef(&git.UpdateRef{
```
jainankit, line 98: To confirm, at this point both headRef and baseRef would be same?

internal/stacks/stacks.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		return nil, errors.WrapIff(
+			err,
+			"failed to create squash commit for branch",
+		)
+	}
+	logrus.WithField("sha", git.ShortSha(squashCommit)).Debug("created synthetic squash commit")
+
+	// Save the squash commit as a new branch.
+	baseRefName := "refs/av/stack-base/" + opts.Name
```
jainankit, line 90: do these act as regular branches, can we create PR with `ref/av/...` as base branch?

cmd/av/main.go
travigd: I wanted it match `git -C ...` which runs Git in another directory than the working directory. The reason it's needed at all is really just during development. I don't want to run experimental Git commands in this repo to test it, but you have to be in the `av` directory for `go run` to work.

So it's usually:
```
$ pwd
~/Aviator/av

$ go run ./cmd/av -C ~/Aviator/test-repo stack branch
...
```

internal/git/git.go
travigd: See comment on `-C` above.

internal/git/git.go
travigd: Maybe!

Could change this to
```
// return the old branch name on success
func (r *Repo) CheckoutBranch(args ...string) (string, error) { ... }
```
and usage to
```
lastBranch, err := repo.CheckoutBranch("-b", "foo")
if err != nil { ... }
defer repo.CheckoutBranch(lastBranch)
```

internal/stacks/stacks.go
travigd: Currently, we only support creating a stack-branch that's twice-removed from `main`.

So if we have `main -> feature1` (`feature1` is just a "normal" branch), we can create `feature1 -> feature2` by squashing all the commits from `feature2`. However, to do it for `feature2 -> feature3`, we need to be able to figure out that `feature2` is stacked on top of `feature1` -- which it can't right now.

I also think that reading all the stackmetadata will be how we construct the tree (i.e., list all the stackmetadata refs and construct a tree out of them).

internal/utils/cleanup/cleanup.go
travigd: It's just a nice utility thing when you're doing something that might need to be cleaned up if the process fails.

So for example, when creating a stack branch, we do a sequence of operations that should be cleaned up if any of the subsequent operations fail. Right now I think it's mostly just reverting the checkout. You could also write it using just defer but it's a little jankier.

```go
// using this Cleanup utility
var cu cleanup.Cleanup
... = repo.Checkout(...)
cu.Add(func() { repo.Checkout(originalBranch) })

// do more stuff

// on success, we cancel the cleanup so they don't run
cu.Cancel()
return ...
```
versus
```go
// using this Cleanup utility
didSucceed := false
... = repo.Checkout(...)
defer func() {
    if didSucceed { return }
    repo.Checkout(originalBranch)
}()

// do more stuff

// on success, we cancel the cleanup so they don't run
didSucceed = true
return ...
```

internal/stacks/stacks.go
travigd: No, branches have to be of the form `refs/heads/...`. We'll need to do that before push push to GitHub, but I'm thinking it'll be nice to have a way to get our version of the base branch (so that we can make sure it hasn't changed remotely).

Ultimately we could possibly fold that into the stack metadata, but that remains to be seen.

internal/stacks/stacks.go
travigd: Yep! Then you create commits on the headRef just like normal when you're working on a feature.

internal/stacks/stacks.go
travigd: With the head branch we just want to make sure we're not clobbering an existing branch. That should matter less here because the name of this ref is exactly determined by the branch name.

jainankit (approved):

PR #5, opened 2022-04-21
[MER-671] Support for abritrary depth stacks

travigd:


internal/stacks/stacks.go
```diff
@@ -23,83 +24,154 @@ type BranchOpts struct {
...
-		return nil, errors.WrapIff(
-			err,
-			"failed to create squash commit for branch",
-		)
+		return errors.WrapIf(err, "failed to render commit message")
 	}
-	logrus.WithField("sha", git.ShortSha(squashCommit)).Debug("created synthetic squash commit")
+	squashCommit, err := repo.Git(
+		"commit-tree",
+		parentHead+"^{tree}", "-p", parentBase,
```
jainankit, line 158: this gives me headache

internal/stacks/stacks.go
```diff
@@ -108,34 +180,41 @@ func CreateBranch(repo *git.Repo, opts *BranchOpts) (*Branch, error) {
 		New: squashCommit,
 		Old: git.Missing,
 	}); err != nil {
-		logrus.WithError(err).Debug("failed to create new branch, it probably already exists")
-		return nil, errors.Errorf("cannot create branch %s (does it already exist?)", opts.Name)
+		return errors.WrapIff(err, "cannot create branch %q", opts.Name)
 	}
 	// Note: this can't include refs/head/... because that enters a detached
```
jainankit, line 185: can you explain this comment more?

jainankit (approved):
Most of this makes sense, let's discuss the metadata in the morning.

internal/stacks/stacks.go
```diff
@@ -5,11 +5,12 @@ import (
...
 	"encoding/json"
 	"github.com/aviator-co/av/internal/git"
-	"github.com/aviator-co/av/internal/utils/cleanup"
+	"github.com/aviator-co/av/internal/utils/templateutils"
 	"github.com/sirupsen/logrus"
+	"text/template"
 )
 
-type Branch struct {
+type branchMetadata struct {
```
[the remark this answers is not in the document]
travigd: In Go, capitalization is the way to say private/public (it's enforced by the compiler). I don't think we need to use this struct outside of this package, so I'd rather it be private.

PR #7, opened 2022-04-21
[MER-670] Implement git.Repo.ListRefs

travigd:
This is preparation for being able to run `av pr` -- we want to be able to figure out what refs already have remote tracking information.

Also added a test case! Yay!

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -9,12 +9,6 @@ import (
 	"strings"
 )
 
-// Missing is a sentinel zero-value for object id (aka sha).
-// Git treats this value as "this thing doesn't exist".
-// For example, when updating a ref, if the old value is specified as EmptyOid,
-// Git will refuse to update the ref if already exists.
-const Missing = "0000000000000000000000000000000000000000"
-
 type Repo struct {
 	repoDir       string
 	defaultBranch string
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -10,45 +10,26 @@ import (
 )
 
 type Repo struct {
-	// immutable
 	repoDir string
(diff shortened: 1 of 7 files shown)


internal/git/listrefs.go
```diff
@@ -0,0 +1,52 @@
+package git
+
+import (
+	"emperror.dev/errors"
+	"strings"
+)
+
+func (r *Repo) ListRefs(showRef *ListRefs) ([]RefInfo, error) {
```
jainankit, line 8: please add some documentation here explaining the purpose.

jainankit (approved):

internal/git/git.go
```diff
@@ -9,36 +9,46 @@ import (
...
 	return r, nil
 }
 
-func (r *Repo) DefaultBranch() string {
-	return r.defaultBranch
+func (r *Repo) DefaultBranch() (string, error) {
+	if r.defaultBranch == "" {
+		// TODO: we should just cache this in some metadata file somewhere
+		// (e.g. .git/av-metadata)
+		refs, err := r.Git("ls-remote", "--symref", "origin", "HEAD")
```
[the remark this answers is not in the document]
travigd: 👏 👏  thank you!

internal/git/listrefs.go
jainankit, line 8: also annoying to have the type-struct same as method name. hard to read.

travigd: Eh it's the best we can do given the lack of support for keyword arguments. The idea is to add ability to specify more fields in the future if necessary. It might not matter a whole lot for this method, but I wanted to be consistent with the `git` API.

PR #8, opened 2022-04-22
[MER-670] Add config file support

travigd:


internal/config/config.go
```diff
@@ -1,5 +1,52 @@
...
+// error if one occurred.
+func Load(paths []string) (bool, error) {
+	config := viper.New()
+
+	// Viper has support for various formats, so it supports kson, toml, yaml,
+	// and more (https://github.com/spf13/viper#reading-config-files).
+	config.SetConfigName("config")
+
+	// Reasonable places to look for config files.
+	config.AddConfigPath("$XDG_CONFIG_HOME/av")
```
jainankit, line 29: what's this? `$XDG_CONFIG_HOME`

jainankit: also this should probably be `.av` as well.

internal/config/config.go
```diff
@@ -1,5 +1,52 @@
...
+
+	if err := config.ReadInConfig(); err != nil {
+		if errors.As(err, &viper.ConfigFileNotFoundError{}) {
+			return false, nil
+		}
+		return false, err
+	}
+
+	if err := config.UnmarshalKey("github", &GitHub); err != nil {
+		return true, errors.Wrap(err, "failed to read github config")
```
jainankit, line 48: so we are saying Github Token is a required field?

internal/config/config.go
travigd: It's just a standard for where people put configuration if it's defined. It shouldn't be `.av` since it's already inside a config directory.

internal/config/config.go
travigd: No. In general, Go has a value of "reasonable zero values", so if it's not defined in the config, it just won't do anything. Essentially the only thing that's not valid would be where the types are wrong:

```yaml
# config.yaml
github:
    token:
        foo: 123
```

jainankit (approved):

PR #10, opened 2022-04-25
Implement av pr create command

travigd:


cmd/av/pr.go
```diff
@@ -18,15 +19,63 @@ var prCreateCmd = &cobra.Command{
...
+		}
+		origin, err := repo.Origin()
+		if err != nil {
+			return errors.WrapIf(err, "failed to determine origin")
+		}
+
+		// figure this out based on whether or not we're on a stacked branch
+		var prBaseBranch string
+
+		stackBranch, err := stacks.GetBranch(repo, currentBranch)
```
jainankit, line 48: prefer: GetBranchMeta. It's not quite clear otherwise why are you passing the branch and getting branch back. Could also be `GetBranchInfo`

cmd/av/pr.go
```diff
@@ -18,15 +19,63 @@ var prCreateCmd = &cobra.Command{
...
+		if err != nil {
+			return errors.WrapIf(err, "failed to get stack branch")
 		}
-		return errors.New("unimplemented")
+		if stackBranch != nil {
+			prBaseBranch = stackBranch.ParentBranchName()
+		} else {
+			defaultBranch, err := repo.DefaultBranch()
```
jainankit, line 55: at this point, this is probably not a stacked branch. Should we share something with the user?

internal/stacks/branch.go
```diff
@@ -0,0 +1,21 @@
...
+import "github.com/aviator-co/av/internal/git"
+
+type Branch struct {
+	metadata *branchMetadata
+}
+
+// GetBranch loads information about the branch from the git repository.
+// Returns nil if given branch does not exist or is not a stacked branch.
+func GetBranch(repo *git.Repo, branchName string) (*Branch, error) {
+	meta := readStackMetadata(repo, branchName)
```
jainankit, line 12: where is this method defined?

internal/git/git.go
```diff
@@ -139,3 +147,34 @@ func (r *Repo) UpdateRef(update *UpdateRef) error {
...
+
+type Origin struct {
+	URL *url.URL
+	// The URL slug that corresponds to repository.
+	// For example, github.com/my-org/my-repo becomes my-org/my-repo.
+	RepoSlug string
+}
+
+func (r *Repo) Origin() (*Origin, error) {
+	// Note: `git remote git-url` gets the "real" URL of the remote (taking
```
jainankit, line 159:
```suggestion
	// Note: `git remote get-url` gets the "real" URL of the remote (taking
```

internal/git/git.go
```diff
@@ -27,7 +29,13 @@ func (r *Repo) DefaultBranch() (string, error) {
 	ref, err := r.Git("symbolic-ref", "refs/remotes/origin/HEAD")
 	if err != nil {
 		logrus.WithError(err).Debug("failed to determine remote HEAD")
-		return "", errors.New("failed to determine remote HEAD (does your repo have a remote configured?)")
+		// this communicates with the remote, so we probably don't want to run
+		// it by default, but we helpfully suggest it to the user. :shrug:
+		logrus.Warn(
+			"Failed to determine repository default branch. " +
+				"Ensure you have a remote named origin and try running `git remote set-head --auto origin` to fix this.",
```
jainankit, line 36: when / how often would this happen?

travigd: Shrug. It happened to me but I couldn't reproduce it after cloning. I think I was mucking around with multiple origins to test something so maybe that was it? I figured it wouldn't hurt to suggest this here.

internal/stacks/branch.go
travigd: It's defined in `stacks.go`.

internal/stacks/create.go
```diff
@@ -0,0 +1,63 @@
...
+	"emperror.dev/errors"
+	"github.com/aviator-co/av/internal/git"
+	"github.com/sirupsen/logrus"
+)
+
+type CreateBranchOpts struct {
+	Name string
+}
+
+// CreateBranch creates a new stack branch based off of the current branch.
```
jainankit, line 13:
```suggestion
// CreateBranch creates and checks out a new stack branch based off of the current branch.
```

internal/stacks/metadata.go
```diff
@@ -0,0 +1,66 @@
...
+	if err := json.Unmarshal([]byte(blob), &branch); err != nil {
+		logrus.WithError(err).WithField("ref", refName).Error("corrupt stack metadata, deleting...")
+		_ = repo.UpdateRef(&git.UpdateRef{Ref: refName, New: git.Missing})
+		return nil
+	}
+
+	return &branch
+}
+
+func writeMetadata(repo *git.Repo, s *BranchMetadata) error {
```
jainankit, line 42: shouldn't this be uppercase if called outside of `metadata.go`

jainankit (approved):

internal/stacks/metadata.go
travigd: No, it's private within the package, not the file.

PR #12, opened 2022-04-26
[MER-689] Implement av stack tree command

travigd:


internal/stacks/stack.go
```diff
@@ -0,0 +1,80 @@
...
+		var meta BranchMetadata
+		meta.Name = strings.TrimPrefix(ref.Revision, "refs/av/stack-metadata/")
+		if err := json.Unmarshal(ref.Contents, &meta); err != nil {
+			return nil, errors.WrapIff(err, "failed to unmarshal metadata for branch %q", ref.Revision)
+		}
+		metas[meta.Name] = &meta
+		hasParent[meta.Name] = true
+		// set default value for parent
+		hasParent[meta.Parent] = hasParent[meta.Parent]
+		children[meta.Parent] = append(children[meta.Parent], meta.Name)
```
jainankit, line 52: something is weird about this whole structure. Are we capturing the scenario where reference has multiple children?

internal/stacks/stack.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	trees := make(map[string]*Tree)
+	// Iterate over all the stack roots and construct a tree for each
+	for branch, hasParent := range hasParent {
```
jainankit, line 57: this is very hard to follow
```suggestion
	for branch, hasParent := range hasParentMap {
```

cmd/av/stack.go
```diff
@@ -91,10 +91,29 @@ var stackTreeCmd = &cobra.Command{
...
+		}
+		for _, tree := range trees {
+			printStackTree(tree, 0)
+		}
+		return nil
 	},
 }
 
+func printStackTree(tree *stacks.Tree, depth int) {
+	indent := strings.Repeat("    ", depth)
```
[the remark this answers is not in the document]
travigd: Tabs are too deep imho.

internal/git/catfile.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+		item.Contents = make([]byte, size)
+		_, err = io.ReadFull(output, item.Contents)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to read cat-file output")
+		}
+
+		// output includes a newline after the item contents
+		// we have to read it otherwise subsequent iterations will fail
+		// (Scanf refuses to consume the newline)
```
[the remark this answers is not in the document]
travigd: Scanf won't advance past newlines unless you include it in the scan format string. So it will fail for `%s %s` if given `hello\nworld`. This comment is just saying we need to read the newline that Git prints at the end of the file contents so that the next Scanf call will work expected (instead of failing since it's trying to read an empty line).

internal/stacks/stack.go
travigd: Yeah. I think some of the complexity is rooted in the fact that we're not (...currently?) tracking any stack metadata for a branch off of trunk. Maybe we just should (we can write that metadata when we're creating the second branch in the stack). I'll take another pass at making this more grok-able in the meantime though.

internal/stacks/stack.go
```diff
@@ -0,0 +1,85 @@
...
+		branchMetadata[branch] = &BranchMetadata{
+			Name: branch,
+		}
+		trees[branch] = completeTree(branch, branchChildren, branchMetadata)
+	}
+
+	return trees, nil
+}
+
+func completeTree(branchName string, branchChildren map[string][]string, branches map[string]*BranchMetadata) *Tree {
```
jainankit, line 75: nit: line too long

cmd/av/stack.go
```diff
@@ -91,10 +91,29 @@ var stackTreeCmd = &cobra.Command{
 	Use:   "tree",
 	Short: "show the tree of stacked branches",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return errors.New("unimplemented")
+		repo, err := getRepo()
+		if err != nil {
+			return err
+		}
+		trees, err := stacks.GetTrees(repo)
```
jainankit, line 98: I'm assuming there's no easy to cache this?

jainankit (approved):

internal/stacks/stack.go
```diff
@@ -0,0 +1,85 @@
...
+			return nil, errors.WrapIff(err, "failed to unmarshal metadata for branch %q", ref.Revision)
+		}
+		branchMetadata[meta.Name] = &meta
+		branchParent[meta.Name] = meta.Parent
+		branchChildren[meta.Parent] = append(branchChildren[meta.Parent], meta.Name)
+	}
+
+	// 2. Find all the branches that do not have a parent. These are the roots.
+	trees := make(map[string]*Tree)
+	for branch := range branchChildren {
```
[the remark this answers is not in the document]
travigd: `range` with one param is keys (or indices). With two params it's the key and the value. Here, I just want to iterate over all the branches with childrens, so I just need the key (which is the branch name).

cmd/av/stack.go
travigd: I don't think it should be too computationally intensive. Think you'd need 1000s of stack branches for it to matter.

Though will need to make sure we're handling "garbage collection" smartly (removing stack metadata for branches that have been merged and whatnot).

PR #13, opened 2022-05-04
Implement stack sync in no-conflict case

travigd:


cmd/av/stack.go
```diff
@@ -83,7 +85,65 @@ stack. This is useful for rebasing a whole stack on the latest changes from the
...
+		repo, err := getRepo()
+		if err != nil {
+			return err
+		}
+		originalBranch, err := repo.CurrentBranchName()
+		if err != nil {
+			return err
+		}
+
+		defer func() {
```
jainankit, line 97: I think for case of sync, let's also validate that there are no uncommitted changes before doing anything else.

cmd/av/stack.go
```diff
@@ -83,7 +85,65 @@ stack. This is useful for rebasing a whole stack on the latest changes from the
...
+			})
+			if err != nil {
+				return errors.WrapIff(err, "failed to sync branch %q", current.Branch.Name)
+			}
+			switch res.Status {
+			case stacks.SyncAlreadyUpToDate:
+				fmt.Printf("Branch %q is already up-to-date with %q\n", current.Branch.Name, current.Branch.Parent)
+			case stacks.SyncUpdated:
+				fmt.Printf("Branch %q synchronized with %q\n", current.Branch.Name, current.Branch.Parent)
+			case stacks.SyncConflict:
```
jainankit, line 132: I'm assuming this is where we have a TODO?

internal/stacks/sync.go
```diff
@@ -0,0 +1,106 @@
...
+	StrategyRebase SyncStrategy = iota
+)
+
+type SyncBranchOpts struct {
+	// The branch to sync with.
+	Parent string
+	// The strategy for performing the sync.
+	// If the branch is already up-to-date, the sync will be a no-op and
+	// strategy will be ignored.
+	Strategy SyncStrategy
```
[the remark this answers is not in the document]
travigd: Sure. Default is `StrategyMergeCommit` because it's the zero value.

cmd/av/stack.go
travigd: Yep. Working on this right now.

cmd/av/stack.go
```diff
@@ -83,7 +85,65 @@ stack. This is useful for rebasing a whole stack on the latest changes from the
...
+			}
+		}()
+
+		root, err := stacks.GetCurrentRoot(repo)
+		if err != nil {
+			return err
+		}
+		printStackTree(root, 0)
+
+		if len(root.Next) == 0 {
```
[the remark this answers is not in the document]
travigd: No, it should never happen™. If you're on main, or on a non-stacked branch, `GetCurrentRoot` will return an error.

jainankit (approved):

PR #14, opened 2022-05-06
[MER-672] Handle stack sync with conflicts

travigd:


internal/git/diff.go
```diff
@@ -32,9 +32,9 @@ func (r *Repo) Diff(d *DiffOpts) (*Diff, error) {
 	contents, err := r.Git(args...)
 	var exitError *exec.ExitError
 	if errors.As(err, &exitError) && exitError.ExitCode() == 1 {
-		return &Diff{Empty: true}, nil
+		return &Diff{Empty: false}, nil
```
travigd, line 35: lol

internal/git/git.go
```diff
@@ -63,6 +68,35 @@ func (r *Repo) Git(args ...string) (string, error) {
...
+	Args []string
+}
+
+type Output struct {
+	ExitCode int
+	Stdout   []byte
+	Stderr   []byte
+}
+
+func (r *Repo) Run(opts *RunOpts) (*Output, error) {
```
travigd, line 81: New Git API because I want it to be more flexible/extensible. In the future will probably try to have the other one be something like `.RunSimple` which just returns a string and error.

cmd/av/stack_sync.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	currentTree := root
+	for {
+		if currentTree.Branch.Name == currentBranch {
+			break
+		}
+		if len(currentTree.Next) > 1 {
```
jainankit, line 171: when would this be a case? I'm assuming we are not going to fail on the condition when root (main) has multiple branches as children which is pretty much always the case. Is current tree representation of root to current branch? If so, when can Next be > 1?

internal/git/git.go
jainankit: So this is same as above `func Git` but have Output object? Technically we can also just use this one everywhere?

jainankit (approved):

internal/git/git.go
travigd: Yeah. I didn't wanna deal with refactoring everything right now just because the diff is already massive. But this is strictly more powerful than `.Git`.

cmd/av/stack_sync.go
```diff
@@ -0,0 +1,280 @@
...
+
+By default, this command will sync all branches starting at the root of the
+stack and recursively rebasing each branch based on the latest commit from the
+parent branch.
+
+If the --current flag is given, this command will not recursively sync dependent
+branches of the current branch within the stack. This allows you to make changes
+to the current branch before syncing the rest of the stack.
+
+If the --trunk flag is given, this command will synchronize changes from the
```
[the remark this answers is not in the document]
travigd: Yeah, I agree (though we can workshop the flag name a bit). Let's do that in a later PR. 😛 (I'm thinking maybe `--latest-trunk` or `--to-trunk` rather than `--with-trunk`)

cmd/av/stack_sync.go
```diff
@@ -0,0 +1,280 @@
...
+
+		// check for --continue/--abort
+		if stackSyncFlags.Continue && stackSyncFlags.Abort {
+			return errors.New("cannot use --continue and --abort together")
+		}
+		existingState, err := readStackSyncState(repo)
+		if err != nil && !os.IsNotExist(err) {
+			return err
+		}
+		if stackSyncFlags.Abort {
```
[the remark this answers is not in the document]
travigd: Yeah, they can do whatever they want. They shouldn't though, and git will yell at them if there are merge conflicts.

TBH I'd like to merge this as-is, then think hard about edge case user experience later.

cmd/av/stack_sync.go
travigd: The trunk branch (`main`/`master`/etc) is never considered a root. The root refers to the first off-trunk branch. So if we have:
```
main
  bookstore-backend
    bookstore-frontend
  petshop-backend
    petshop-frontend
```
the stack roots are `bookstore-backend` and `petshop-backend`.

cmd/av/stack_sync.go
```diff
@@ -0,0 +1,280 @@
...
+		if len(currentTree.Next) > 1 {
+			return false, errors.Errorf("unsupported: branch %q has multiple stack children", currentTree.Branch.Name)
+		}
+		if len(currentTree.Next) == 0 {
+			return false, errors.Errorf("invariant error: couldn't find branch %q in stack", currentBranch)
+		}
+		currentTree = currentTree.Next[0]
+	}
+
+	for {
```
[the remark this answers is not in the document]
travigd: No, this whole section of code is... awkward at the moment. Which is a big part of the reason I want to re-work how we store branch metadata.

The current code API gives us the root, then we have to traverse the root to find the current branch (which is the first for loop) and then continue synchronizing every branch down the tree from that point (the second loop).

I'd like to take a stab at refactoring this after #15 merges as well.

cmd/av/stack_sync.go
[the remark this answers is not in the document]
jainankit: Sure we can workshop the name, but I think is syncing from trunk and not to trunk.

cmd/av/stack_sync.go
jainankit: technically, even in that case, we should not have raise an error like this. So if you are doing sync in such as  setup:
```
main
   first
     first-x
     first-y
        first-y-1
        first-y-2 *
  
```
and you can av stack sync from `first-y-2`, we should sync `first` to `first-y` and `first-y` to `first-y-2`.

PR #15, opened 2022-05-10
[MER-758] Store PR metadata locally in the repo

travigd:
Created this PR using the `av pr create` command. ;)


cmd/av/init.go
```diff
@@ -0,0 +1,64 @@
...
+	"context"
+	"emperror.dev/errors"
+	"fmt"
+	"github.com/aviator-co/av/internal/config"
+	"github.com/aviator-co/av/internal/gh"
+	"github.com/aviator-co/av/internal/meta"
+	"github.com/spf13/cobra"
+)
+
+var initFlags struct {
```
travigd, line 13: I don't love having an init command, but I think it's fine for now.

cmd/av/pr.go
```diff
@@ -1,53 +1,97 @@
...
+			}
+		}
+
+		// TODO:
+		//     It would be nice to be able to auto-detect that a PR has been
+		//     opened for a given PR without using av. We might need to do this
+		//     when creating PRs for a whole stack (e.g., when running `av pr`
+		//     on stack branch 3, we should make sure PRs exist for 1 and 2).
+		branchMeta, ok := meta.GetBranch(repo, currentBranch)
+		if ok && branchMeta.PullRequest.ID != "" && !prCreateFlags.Force {
```
jainankit, line 87: I'm assuming the GetBranch in future will populate `PullRequest.ID`?

cmd/av/pr.go
```diff
@@ -1,53 +1,97 @@
...
 import (
+	"context"
 	"emperror.dev/errors"
 	"fmt"
 	"github.com/aviator-co/av/internal/config"
-	"github.com/aviator-co/av/internal/stacks"
+	"github.com/aviator-co/av/internal/gh"
+	"github.com/aviator-co/av/internal/git"
+	"github.com/aviator-co/av/internal/meta"
+	"github.com/shurcooL/githubv4"
```
jainankit, line 11: how legit is this? Do we need to fork it?

cmd/av/pr.go
```diff
@@ -84,4 +155,24 @@ func init() {
 		&prCreateFlags.Base, "base", "",
 		"base branch to create the pull request against",
 	)
+	prCreateCmd.Flags().BoolVar(
+		&prCreateFlags.Force, "force", false,
+		"force creation of a pull request even if one already exists",
```
jainankit, line 160:
```suggestion
		"force create a pull request even if one already exists",
```

cmd/av/pr.go
```diff
@@ -63,15 +107,42 @@ var prCreateCmd = &cobra.Command{
...
-			currentBranch,
-		)
+		client, err := gh.NewClient(config.GitHub.Token)
+		if err != nil {
+			return err
+		}
+
+		body := prCreateFlags.Body
+		// Special case: ready body from stdin
+		if body == "-" {
```
jainankit, line 117: so empty body in params defaults to a `-`? Shouldn't this be an empty string

cmd/av/sync.go
```diff
@@ -0,0 +1,40 @@
...
+var syncCmd = &cobra.Command{
+	Use:   "sync",
+	Short: "fetch latest information from GitHub",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		client, err := gh.NewClient(config.GitHub.Token)
+		if err != nil {
+			return errors.Wrap(err, "failed to create GitHub client")
+		}
+		pull, err := client.PullRequest(context.Background(), gh.PullRequestOpts{
+			Owner:  "travigd",
```
jainankit, line 22: remove hard-coded values

cmd/av/sync.go
```diff
@@ -0,0 +1,40 @@
...
+			Repo:   "private-playground",
+			Number: 1,
+		})
+		if err != nil {
+			return errors.WrapIf(err, "failed to fetch pull request")
+		}
+		data, err := json.MarshalIndent(pull, "", "  ")
```
jainankit, line 29: what happens if pull doesn't exist, do we get an error?

internal/meta/repository.go
```diff
@@ -0,0 +1,50 @@
...
+type Repository struct {
+	// The GitHub (GraphQL) ID of the repository (e.g., R_kgDOHMmHmg).
+	ID string `json:"id"`
+	// The owner of the repository (e.g., aviator-co)
+	Owner string `json:"owner"`
+	// The name of the repository (e.g., av)
+	Name string `json:"name"`
+}
+
+func GetRepository(repo *git.Repo) (Repository, bool) {
```
jainankit, line 22: prefer if we switch to Read* and Write* for meta, vs Get* and Create* for API calls and other actions.

jainankit (approved):

cmd/av/pr.go
```diff
@@ -1,53 +1,97 @@
...
+			}
+			logrus.WithField("upstream", upstream).Debug("pushing latest changes")
+			if _, err := repo.Git("push"); err != nil {
+				return errors.WrapIf(err, "failed to push")
+			}
+		}
+
+		// TODO:
+		//     It would be nice to be able to auto-detect that a PR has been
+		//     opened for a given PR without using av. We might need to do this
```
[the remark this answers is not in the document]
travigd: 💯

cmd/av/pr.go
travigd: `meta.GetBranch` just loads the stored metadata. It already populates `PullRequest.ID` if it's been set (which is what we're checking here).

cmd/av/pr.go
travigd: No, default *is* an empty string, where we just make the body empty. In the future, I want to automatically set it based on the first commit message. But as a special case, you can read from stdin.

So:
```
# Create PR with empty body:
av pr create -t 'Hello, world!'

# Create PR with body from stdin
echo 'This is a cool PR!' | av pr create -t 'Hello, world!' -b -

# Or with heredoc syntax
av pr create -t 'Hello, world!' -b - <<EOF
This is a cool PR!
EOF
```

PR #17, opened 2022-05-10
[MER-762] Rework stack branch metadata

travigd:
Also need to rework the stack tree command, but will do that in a followup PR.

I think this will make navigating the stacks programatically a lot easier.
We'll just have a map/dict of stack metadatas and can move through the stack like:
```
current = branches[current.Parent]
```


cmd/av/stack_branch.go
```diff
@@ -0,0 +1,107 @@
...
+		defer cu.Cleanup()
+
+		repo, err := getRepo()
+		if err != nil {
+			return err
+		}
+
+		name := args[0]
+
+		// validate operation preconditions
```
travigd, line 36: Inlined what used to be `stacks.CreateBranch` here. I think the whole `stacks` package was a little bit of a premature abstraction.

cmd/av/stack_branch.go
```diff
@@ -0,0 +1,107 @@
...
+			Parent: parentBranch,
+		}
+		logrus.WithField("meta", branchMeta).Debug("writing branch metadata")
+		if err := meta.WriteBranch(repo, branchMeta); err != nil {
+			return errors.WrapIff(err, "failed to write av internal metadata for branch %q", name)
+		}
+
+		if parentBranch != "" {
+			parentMeta, _ := meta.ReadBranch(repo, parentBranch)
+			parentMeta.Children = append(parentMeta.Children, name)
```
travigd, line 97: Now we store the children of each branch internally.

cmd/av/stack_next.go
```diff
@@ -0,0 +1,44 @@
+package main
```
travigd, line 1: unchanged, just moved to a new file

cmd/av/stack_prev.go
```diff
@@ -0,0 +1,31 @@
+package main
```
travigd, line 1: unchanged, just moved to a new file

cmd/av/stack_tree.go
```diff
@@ -0,0 +1,35 @@
+package main
```
travigd, line 1: Unchanged, but will address in a followup PR.

internal/meta/branch.go
```diff
@@ -13,15 +13,28 @@ type Branch struct {
...
+	// trunk will have no parent, but we usually don't explicitly consider it a
+	// stack unless it also has stack children)
 	Parent string `json:"parent"`
 
-	PullRequest PullRequest `json:"pullRequest,omitempty"`
+	// The children branches of this branch within the stack (if any).
+	Children []string `json:"children,omitempty"`
+
+	// The associated pull request information, if any.
+	PullRequest *PullRequest `json:"pullRequest,omitempty"`
```
travigd, line 29: Changed to pointer to allow `omitempty` to work correctly.

cmd/av/stack_branch.go
```diff
@@ -0,0 +1,107 @@
...
+			parentBranch, err = repo.CurrentBranchName()
+			if err != nil {
+				return errors.WrapIff(err, "failed to get current branch name")
+			}
+		}
+
+		// Special case: branching from the repository default branch
+		if parentBranch == defaultBranch {
+			logrus.Debug("creating new stack root branch from default branch")
+			parentBranch = ""
```
travigd, line 71: Context around why I'm doing this the way I am:

There were really two options:
* Stack roots (which, remember, are the first non-trunk branch in a stack) have `Parent: ""`
* Stack roots have `Parent: "<default branch>"`

I opted for the former that way when we get metadata for a branch that wasn't created with `av` (e.g., `git checkout -b ...` directly), we just assume it's a stack root branch, which is pretty logical.

jainankit: wait, so if someone creates a branch directly, there's no way to get that into the stack?

cmd/av/stack_branch.go
```diff
@@ -0,0 +1,107 @@
...
+		} else {
+			var err error
+			parentBranch, err = repo.CurrentBranchName()
+			if err != nil {
+				return errors.WrapIff(err, "failed to get current branch name")
+			}
+		}
+
+		// Special case: branching from the repository default branch
+		if parentBranch == defaultBranch {
```
jainankit, line 69: can you document in code this behavior - we are intentionally setting the parent Branch for root branch as empty, so that we can uniquely identify a tree.

jainankit (approved):

PR #23, opened 2022-05-31
[MER-848] Add automatic version check

travigd:
image

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -9,6 +9,8 @@ builds:
     goarch:
       - amd64
       - arm64
+    ldflags:
+      - "-X 'github.com/aviator-co/av/internal/config.Version={{ .Version }}'"
 
 # Create a GitHub release on the av repo
 release: {}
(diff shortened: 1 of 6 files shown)


.goreleaser.yaml
```diff
@@ -28,7 +30,7 @@ checksum:
   name_template: 'checksums.txt'
 
 snapshot:
-  name_template: "{{ incpatch .Version }}-next"
+  name_template: "{{ incpatch .Version }}-dev-{{ .ShortCommit }}"
```
jainankit, line 33: how is this snapshot used, are you sure you want to hard-code dev here?

jainankit (approved):

.goreleaser.yaml
travigd: Yeah, this becomes the version strung but only if you do `goreleaser --snapshot` to generate a snapshot build. When you actually release, it doesn't use this, it uses the actual version string (`v1.2.3` or whatever)

PR #25, opened 2022-06-02
Implement av stack reparent command

travigd:
I originally wanted to do `av stack sync --parent <newparent>`, but it was different enough from a sync that I made it its own command. I'm still not 100% set on that but... 🤷.

Also, it's possible for there to be a rebase conflict during this. It's possible to get into an undesirable state if you then abort that rebase (because av thinks the parent is `<newparent>` but the git branch is still based on `<oldparent>`).

Nonetheless, I think I'd like to merge this as is and think about that limitation a bit later.

### DIFF
--- a/cmd/av/stack_reparent.go
+++ b/cmd/av/stack_reparent.go
@@ -1,0 +1,100 @@
+package main
+
+import (
+	"emperror.dev/errors"
+	"github.com/aviator-co/av/internal/git"
+	"github.com/aviator-co/av/internal/meta"
+	"github.com/aviator-co/av/internal/utils/sliceutils"
+	"github.com/spf13/cobra"
+	"strings"
+)
+
+var stackReparentCmd = &cobra.Command{
+	Use:   "reparent <new-parent>",
+	Short: "change the stack parent of the current branch",
+	Long: strings.TrimSpace(`
+Change the stack parent of the current branch.
+
+This command can be used to add and/or remove a branch from a stack.
+`),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) != 1 {
+			_ = cmd.Usage()
+			return errors.New("expected exactly one argument")
+		}
+		newParent := args[0]
+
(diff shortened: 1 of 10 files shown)


cmd/av/stack_reparent.go
```diff
@@ -0,0 +1,100 @@
...
+			return errors.Errorf("parent branch %q does not exist", newParent)
+		}
+
+		branchMeta, _ := meta.ReadBranch(repo, branch)
+
+		// We might need to rebase the branch on top of the new parent. This
+		// requires a special rebase command because the "normal" rebase command
+		// (without --onto) will consider every commit in the current branch when
+		// figuring out what to be played on top of the new parent. So, for example,
+		// if we have a stack that looks like B1->B2->B3 with corresponding commits
```
jainankit, line 58: This seems like a very weird behavior to me. I'm not sure what someone is thinking when they want to reparent B3 onto B1. If I was running this command, I would assume that I just somehow want to remove the branch B2. But in that case, I would actually still keep C2 and C3 in B3. Trying to "delete" a commit from the tree seems like an odd behavior

jainankit: FWIW, I'm still not certain what's the expected use case of this command. The only few I think of are:
1- I create a branch on main, but I realized I want to actually base it on another branch, so I rebase it there.
2- I created a commit in this stack, but i actually want to move it to main instead.

The second scenario should also cover where we merged all the branches up until that point. So now my branch is pointing to a base branch that's already merged. So reparenting makes sense.

Everything other use case seems like unnecessary complexity.

travigd: I don't really consider wanting to move a branch to a different stack as especially "edge-case-y" behavior.

But, any rate, we still need all of this logic to handle reparenting `3` onto `main` after merging `2` (otherwise it would include the commits from `1` and `2` and try to replay them on top of `main`, which would be a likely merge conflict).

jainankit (approved):

PR #26, opened 2022-06-03
[MER-866] Implement av fetch command

travigd:


cmd/av/debug.go
```diff
@@ -0,0 +1,38 @@
...
+	"context"
+	"fmt"
+	"github.com/aviator-co/av/internal/config"
+	"github.com/aviator-co/av/internal/gh"
+	"github.com/shurcooL/githubv4"
+	"github.com/spf13/cobra"
+)
+
+var debugCmd = &cobra.Command{
+	Use: "debug",
```
jainankit, line 13: No help doc for user?

cmd/av/fetch.go
```diff
@@ -0,0 +1,115 @@
...
+					" open pull requests from GitHub...",
+					"\n",
+				)
+			}
+
+			for _, pr := range prsPage.PullRequests {
+				// TODO: maybe warn if local branch is not up-to-date with remote?
+				branchMeta, ok := branches[pr.HeadBranchName()]
+				if !ok {
+					logrus.WithField("branch", pr.HeadBranchName()).Debug("skipping PR for unknown local branch")
```
jainankit, line 62: suggest them that they should run git fetch?

cmd/av/fetch.go
```diff
@@ -0,0 +1,115 @@
...
+					continue
+				}
+				logrus.WithField("branch", pr.HeadBranchName()).Debug("found PR for known local branch")
+				if branchMeta.PullRequest == nil {
+					_, _ = fmt.Fprint(
```
jainankit, line 67: shouldn't this just be debug info?

jainankit (approved):

cmd/av/fetch.go
travigd: This is just so that we have all of the PR information for any branches you have locally that are being "tracked" by `av`. It's pretty likely that there will be PRs for branches that are not being locally tracked by `av` (e.g., because they are other people's PRs). Running `git fetch` wouldn't change that, because we still wouldn't have `av` metadata for the branch.

cmd/av/fetch.go
travigd: Eh, I'd rather have it print out what it's doing. It's not likely to be too terribly much information.

PR #27, opened 2022-06-03
[MER-867] Add default PR title from commit subject

travigd:
This also moves a lot of the code that was in `./cmd/av/pr.go` into a new folder:
`./internal/actions`. The reason for this is to be able to extract out certain
things that we'll need to do in multiple contexts (in this case, I want to be
able to implement an `av stack submit` command, which will involve opening a PR
for every branch in the stack).

That code is mostly unchanged, except for the bit where we now read the subject
and body from the commit in order to populate the PR title/body. I'm going to
submit this PR using this strategy, so... wish me luck! 🤞

### DIFF
--- a/cmd/av/helpers.go
+++ b/cmd/av/helpers.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"emperror.dev/errors"
 	"github.com/aviator-co/av/internal/git"
 	"github.com/aviator-co/av/internal/meta"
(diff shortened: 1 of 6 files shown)


travigd:
^ It worked!

The newlines on GitHub are a touch weird, but probably not blocking.

jainankit (approved):
🤷

PR #28, opened 2022-06-03
[MER-870] Fix bug in av pr create if commit only has headline (no body)

travigd:


internal/git/show.go
```diff
@@ -19,12 +20,16 @@ type CommitInfo struct {
 func (r *Repo) CommitInfo(opts CommitInfoOpts) (*CommitInfo, error) {
 	// Need --quiet to suppress the diff that would otherwise be printed at the
 	// end
-	res, err := r.Git("show", "--quiet", "--format=%H%n%s%n%b", opts.Rev)
+	res, err := r.Run(&RunOpts{
```
travigd, line 23: the bug was that `r.Git` trims the whitespace from the string, so it would fail to parse if commit body was empty (since the format string is `<hash>\n<subject>\n<body>` and body was empty and the `\n` after subject would be trimmed off)

jainankit (approved):

doratzeng (approved):

PR #33, opened 2022-06-17
[MER-874] Auto Open PR In Browser

alec-stew-art:
added edited code from Go standard library that allows us to open browser on PR creation.

# Testing

- ran `av init` locally on this repository
- generated a test PR on a new branch
- confirmed PR opens up default browser on PR creation for mac running apple silicon
**TODO**: confirm functionality on other operating systems

# Results
Screen Shot 2022-06-17 at 1 48 26 PM

On run of the above command the following window was opened directly in browser
Screen Shot 2022-06-17 at 1 48 13 PM

# Other
Might want to enhance this before merging to allow for the auto-open to not happen if someone doesn't want it to happen.
Maybe a config that is on by default but you can turn off in the `config.yaml`?

### DIFF
--- a/internal/utils/browser/browser.go
+++ b/internal/utils/browser/browser.go
@@ -1,0 +1,62 @@
+package browser
+
+import (
+	"os"
+	"os/exec"
+	"runtime"
+	"time"
+)
+
+// Commands returns a list of possible commands to use to open a url.
+func Commands() [][]string {
+	var cmds [][]string
+	if exe := os.Getenv("BROWSER"); exe != "" {
+		cmds = append(cmds, []string{exe})
+	}
+	switch runtime.GOOS {
+	case "darwin":
+		cmds = append(cmds, []string{"/usr/bin/open"})
+	case "windows":
+		cmds = append(cmds, []string{"cmd", "/c", "start"})
+	default:
+		if os.Getenv("DISPLAY") != "" {
+			// xdg-open is only for use in a desktop environment.
+			cmds = append(cmds, []string{"xdg-open"})
+		}
+	}
+	cmds = append(cmds,
+		[]string{"chrome"},
+		[]string{"google-chrome"},
+		[]string{"chromium"},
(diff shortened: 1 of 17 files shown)


alec-stew-art:
I am not really sure if I have imported anything correctly, and also I still need to do some reading on my end to understand how to locally test the application, but I think something like this general approach works?

internal/utils/browser/browser.go
```diff
@@ -0,0 +1,62 @@
...
+		[]string{"firefox"},
+	)
+	return cmds
+}
+
+// Open tries to open url in a browser and reports whether it succeeded.
+func Open(url string) bool {
+	for _, args := range Commands() {
+		cmd := exec.Command(args[0], append(args[1:], url)...)
+		if cmd.Start() == nil && appearsSuccessful(cmd, 3*time.Second) {
```
travigd, line 40: We don't need `appearsSuccessful` below -- that'll just introduce extra latency in a way that I don't think matters for us. Just change to `if err := cmd.Start(); err != nil { ... }`.

travigd: Actually, this function should probably have signature `func Open(url string) error` where we just return the (wrapped) error from `cmd.Start`.

alec-stew-art:
Retested with code changes and we still open a browser window successfully - haven't tested failure message yet but should be fine.

Screen Shot 2022-06-23 at 12 06 31 PM

internal/utils/browser/browser.go
```diff
@@ -1,62 +1,64 @@
...
 	"runtime"
-	"time"
 )
 
-// Commands returns a list of possible commands to use to open a url.
-func Commands() [][]string {
-	var cmds [][]string
+// Open tries to open url in a browser and reports whether it succeeded.
+func Open(url string) error {
+	var args []string = nil
```
travigd, line 40: nit: don't need `= nil` here

travigd (approved):

PR #36, opened 2022-06-23
[MER-927] auto open browser config

alec-stew-art:
Added top level config value for controlling if the auto-open browser behavior from MER-874

# Test Plan
- add `openbrowser` false to `~/.av/config.yaml`
- open new PR on test repo and see that browser now does not open

# Results
Added `openbrowser: false` to config file and ran the following which -> browser did not open:

Screen Shot 2022-06-23 at 2 11 57 PM

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -5,17 +5,19 @@ import (
 	"github.com/spf13/viper"
 )
 
-var GitHub = struct {
+type GitHub struct {
 	Token   string
 	BaseUrl string
-}{
-	BaseUrl: "https://github.com",
 }
 
-var App = struct {
+var Av = struct {
 	OpenBrowser bool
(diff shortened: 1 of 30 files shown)


internal/config/config.go
```diff
@@ -43,8 +51,8 @@ func Load(paths []string) (bool, error) {
 		return false, err
 	}
 
-	if err := config.UnmarshalKey("github", &GitHub); err != nil {
-		return true, errors.Wrap(err, "failed to read github config")
+	if err := config.Unmarshal(&Av); err != nil {
+		return true, errors.Wrap(err, "failed to read av configs")
```
travigd, line 55:
```suggestion
		return true, errors.Wrap(err, "failed to read av configuration")
```

travigd (approved):

PR #39, opened 2022-06-30
[MER-873] Highlight current branch in the stack

jainankit:
Screen Shot 2022-06-30 at 1 14 43 PM

### DIFF
--- a/cmd/av/stack_tree.go
+++ b/cmd/av/stack_tree.go
@@ -2,9 +2,13 @@ package main
 
 import (
 	"fmt"
-	"github.com/aviator-co/av/internal/meta"
-	"github.com/spf13/cobra"
+	"os"
 	"strings"
+
+	"github.com/aviator-co/av/internal/git"
+	"github.com/aviator-co/av/internal/meta"
+	"github.com/aviator-co/av/internal/utils/colors"
+	"github.com/spf13/cobra"
 )
 
 var stackTreeCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


cmd/av/stack_tree.go
```diff
@@ -26,32 +30,49 @@ var stackTreeCmd = &cobra.Command{
...
+			return err
+		}
+
 		// TODO[polish]:
 		// 		We should show information about whether or not each branch is
 		//     	up-to-date with its stack parent as well as whether or not the
 		//		branch is up-to-date with its upstream tracking branch.
-		fmt.Println(defaultBranch)
+		if currentBranch == defaultBranch {
+			_, _ = fmt.Fprint(os.Stderr,
```
travigd, line 43: We're using `fmt.Println` below which prints to stdout, not stderr, so let's be consistent here.

```suggestion
			_, _ = fmt.Print(
```

cmd/av/stack_tree.go
```diff
@@ -26,32 +30,49 @@ var stackTreeCmd = &cobra.Command{
...
-			printStackTree(branches, branch, 1)
+			printStackTree(branches, repo, currentBranch, branch, 1)
 		}
 
 		return nil
 	},
 }
 
-func printStackTree(branches map[string]meta.Branch, root string, depth int) {
+func printStackTree(branches map[string]meta.Branch, repo *git.Repo, currentBranch string, root string, depth int) {
```
travigd, line 61: nit: I think repo should be the first arg here (so it should be `repo, branches, currentBranch, root, depth` respectively).

travigd (approved):

PR #47, opened 2022-07-18
fix(stack sync): Handle stack sync rebase for >= 3 branches correctly

travigd:


alec-stew-art (approved):
Went over everything and testing/name changes looks good.
The new way of managing the "root" branch with trunk seems to also be a lot cleaner.

PR #48, opened 2022-07-22
fix(stack-sync): cleanup rebase state on sync --abort

travigd:


cmd/av/stack_sync.go
```diff
@@ -122,13 +114,32 @@ base branch.
 			if state.CurrentBranch == "" {
 				return errors.New("no sync in progress")
 			}
+
+			// Abort the rebase if we need to
```
jainankit, line 118: so previously the abort was a no-op? We were only clearing the state from the refs without actually updating any references?

jainankit (approved):

cmd/av/stack_sync.go
travigd: Yep, that's correct. 🙃 Kind of an obvious oversight in retrospect but hindsight is 20/20 😅

PR #49, opened 2022-08-01
feat(stacks): added shortcut flag to stack prev and stack next

alec-stew-art:
# What changed
added `--first` and `--last` shortcuts to `av stack prev` and `av stack next` respectively.

# Testing
Tested locally creating a stack of branches `blah1`, `blah2`, and finally `blah3` and then ran the following commands to show things working:
Screen Shot 2022-08-01 at 4 11 00 PM

### DIFF
--- a/cmd/av/stack_next.go
+++ b/cmd/av/stack_next.go
@@ -18,23 +18,12 @@ var stackNextFlags struct {
 }
 
 var stackNextCmd = &cobra.Command{
-	Use:   "next <n> or next --last",
+	Use:   "next [<n>|--last]",
 	Short: "checkout the next branch in the stack",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		var n int = 1
-		if len(args) == 1 && !stackNextFlags.Last {
-			var err error
-			n, err = strconv.Atoi(args[0])
-			if err != nil {
-				return errors.New("invalid number")
-			}
-		} else if len(args) > 1 {
+		if len(args) > 1 {
 			_ = cmd.Usage()
 			return errors.New("too many arguments")
-		}
-
-		if n <= 0 {
-			return errors.New("invalid number (must be >= 1)")
 		}
 
 		// Get the subsequent branches so we can checkout the nth one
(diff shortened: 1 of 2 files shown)


cmd/av/stack_prev.go
```diff
@@ -12,12 +12,17 @@ import (
...
 )
 
+var stackPrevFlags struct {
+	// should we go to the first
+	First bool
+}
+
 var stackPrevCmd = &cobra.Command{
-	Use:   "prev <n>",
+	Use:   "prev <n> or prev --first",
```
travigd, line 21:
```suggestion
	Use:   "prev [<n>|--first]",
```

cmd/av/stack_prev.go
```diff
@@ -12,12 +12,17 @@ import (
...
 	Short: "checkout the previous branch in the stack",
 	RunE: func(cmd *cobra.Command, args []string) error {
 		var n int = 1
-		if len(args) == 1 {
+		if len(args) == 1 && !stackPrevFlags.First {
```
travigd, line 25: Again, I think it would be better to just determine the target branch as two distinct branches rather than having them be interlocked like this.

cmd/av/stack_next.go
```diff
@@ -12,12 +12,17 @@ import (
...
 )
 
+var stackNextFlags struct {
+	// should we go to the last
+	Last bool
+}
+
 var stackNextCmd = &cobra.Command{
-	Use:   "next <n>",
+	Use:   "next <n> or next --last",
```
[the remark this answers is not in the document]
travigd: FYI I copied this syntax from how Git tends to do it.

cmd/av/stack_next.go
```diff
@@ -50,16 +44,29 @@ var stackNextCmd = &cobra.Command{
...
 		}
 
-		// confirm we can in fact do the operation given current branch state
-		if len(subsequentBranches) == 0 {
-			return errors.New("there is no next branch")
-		}
-		if n > len(subsequentBranches) {
-			return fmt.Errorf("invalid number (there are only %d subsequent branches in the stack)", len(subsequentBranches))
+		var branchToCheckout string
+		if !stackNextFlags.Last {
```
travigd, line 48: Nit: let's make the non-negated condition first.

cmd/av/stack_next.go
```diff
@@ -50,16 +44,29 @@ var stackNextCmd = &cobra.Command{
...
+			}
+			if n <= 0 {
+				return errors.New("invalid number (must be >= 1)")
+			}
+			if n > len(subsequentBranches) {
+				return fmt.Errorf("invalid number (there are only %d subsequent branches in the stack)", len(subsequentBranches))
+			}
+			branchToCheckout = subsequentBranches[n-1]
+		} else {
+			if len(subsequentBranches) == 0 {
```
travigd, line 64: Let's skip this check. Or at the very least, just print out a message and return `nil` here. I don't think it properly counts as an **error**.

cmd/av/stack_next.go
```diff
@@ -50,16 +44,29 @@ var stackNextCmd = &cobra.Command{
...
-			return errors.New("there is no next branch")
-		}
-		if n > len(subsequentBranches) {
-			return fmt.Errorf("invalid number (there are only %d subsequent branches in the stack)", len(subsequentBranches))
+		var branchToCheckout string
+		if !stackNextFlags.Last {
+			if len(subsequentBranches) == 0 {
+				return errors.New("there is no next branch")
+			}
+			n, err := strconv.Atoi(args[0])
```
travigd, line 52: You need to check the length of args here. Can probably just skip the `len(args)` check at the top and inline it here (and make sure `len(args)` is zero if `--last` is given).

So something like (in atrocious pseudocode):

```
if --last {
  assert len(args) == 0
  targetBranch = ...
} else {
  n = 1
  if len(args) == 1 {
    n = parseInt(args[0])
  } else if len(args) > 1 {
    return error
  }
}
```

travigd (approved):
Gonna approve this, should just take one last minor revision. I didn't look at `prev` since I assume it's just mirroring `next`.

travigd:
ALSO if you wanna write a e2e type test (maybe stacked on top of this) that'd be super dope™.

PR #50, opened 2022-08-01
feat(stacks): add av stack submit command

alec-stew-art:
# What Changed
`av stack submit` command was added that pushed the current state of the stack - i.e. creates PRs for every branch that doesn't have them and fixes the base branch of the PRs that already exist to be the correct one.

This covers work for both MER-1024 and MER-1007

# Testing
- created fresh stack locally and ran the command correctly seeing the full stack get submitted
- created fresh stack - create one PR and set the base ref branch to the wrong thing - correctly see the stack submitted and the one wrong base branch get updated.

# Results
Initial PR created for branch `submit3` onto `main` as normal:
Screen Shot 2022-08-03 at 4 09 00 PM
`av stack submit` command called and the three PRs are submitted and in the case of `submit3` corrected:
Screen Shot 2022-08-03 at 4 09 39 PM
Screen Shot 2022-08-03 at 4 09 44 PM
Screen Shot 2022-08-03 at 4 46 03 PM

### DIFF
--- a/cmd/av/stack_submit.go
+++ b/cmd/av/stack_submit.go
@@ -14,7 +14,7 @@ import (
 
 var stackSubmitCmd = &cobra.Command{
 	Use:   "submit",
-	Short: "create pull requests for the current stack",
+	Short: "create/synchronize pull requests for the current stack",
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if len(args) > 0 {
 			_ = cmd.Usage()
@@ -42,13 +42,7 @@ var stackSubmitCmd = &cobra.Command{
 				ctx, repo, client,
 				actions.CreatePullRequestOpts{
 					BranchName: currentMeta.Name,
-					// TODO: currently just set things as empty but it might be nice to have an interactive mode
-					//	that would let us set the title and body, etc. per branch in the stack
-					Title:  "",
-					Body:   "",
-					NoPush: false,
-					Force:  false,
-					Draft:  false,
+					Draft:      config.Av.PullRequest.Draft,
 				},
 			)
 			if err != nil {
--- a/cmd/av/stack_submit.go
+++ b/cmd/av/stack_submit.go
@@ -30,6 +30,22 @@ var stackSubmitCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
+		currentBranch, err := repo.CurrentBranchName()
+		if err != nil {
+			return err
+		}
+		previousBranches, err := meta.PreviousBranches(branches, currentBranch)
+		if err != nil {
+			return err
+		}
+		subsequentBranches, err := meta.SubsequentBranches(branches, currentBranch)
+		if err != nil {
+			return err
+		}
+		var branchesToSubmit []string
+		branchesToSubmit = append(branchesToSubmit, previousBranches...)
+		branchesToSubmit = append(branchesToSubmit, currentBranch)
+		branchesToSubmit = append(branchesToSubmit, subsequentBranches...)
 
 		// ensure pull requests for each branch in the stack
 		ctx := context.Background()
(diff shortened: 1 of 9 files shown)


cmd/av/stack_submit.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/aviator-co/av/internal/config"
+	"github.com/aviator-co/av/internal/gh"
+	"github.com/aviator-co/av/internal/meta"
+	"github.com/shurcooL/githubv4"
+	"github.com/spf13/cobra"
+)
+
+var stackSubmitCmd = &cobra.Command{
+	Use:   "submit",
+	Short: "create pull requests for the current stack",
```
travigd, line 17:
```suggestion
	Short: "create/synchronize pull requests for the current stack",
```

cmd/av/stack_submit.go
```diff
@@ -0,0 +1,72 @@
...
+			_ = cmd.Usage()
+			return errors.New("this command takes no arguments")
+		}
+
+		// Get the all branches in the stack
+		repo, _, err := getRepoInfo()
+		if err != nil {
+			return err
+		}
+		branches, err := meta.ReadAllBranches(repo)
```
travigd, line 29: `ReadAllBranches` returns every branch that `av` knows about, not just the branches in the current stack.

cmd/av/stack_submit.go
```diff
@@ -0,0 +1,72 @@
...
+			return err
+		}
+		for _, currentMeta := range branches {
+			result, err := actions.CreatePullRequest(
+				ctx, repo, client,
+				actions.CreatePullRequestOpts{
+					BranchName: currentMeta.Name,
+					// TODO: currently just set things as empty but it might be nice to have an interactive mode
+					//	that would let us set the title and body, etc. per branch in the stack
+					Title:  "",
```
travigd, line 47: Generally you should just leave these values unspecified (they default to the "zero value" of the type, so an empty string or false).

cmd/av/stack_submit.go
```diff
@@ -0,0 +1,72 @@
...
+					Body:   "",
+					NoPush: false,
+					Force:  false,
+					Draft:  false,
```
travigd, line 51: Use the value from config here.

internal/actions/pr.go
```diff
@@ -135,11 +141,14 @@ func CreatePullRequest(ctx context.Context, repo *git.Repo, client *gh.Client, o
 
 	pull, didCreatePR, err := getOrCreatePR(ctx, client, repoMeta, getOrCreatePROpts{
 		baseRefName: prBaseBranch,
-		headRefName: currentBranch,
+		headRefName: opts.BranchName,
 		title:       opts.Title,
 		body:        opts.Body,
 		draft:       opts.Draft,
 	})
+	if err != nil {
```
travigd, line 149: 🙀 nice catch

travigd: We should add go linting to our CI. It can catch this kind of thing (unchecked `err`).

travigd (commented):
Looks good, just need to change the `ReadAllBranches` so that it only returns the current stack. You can probably create a new function in `meta` that will return a (linearized) stack.

Something like

```go
// GetLinearStack returns the linear ordering of the stack that the target belongs to.
// The target does not need to be the root of the stack.
func GetLinearStack(branches map[string]*meta.Branch, target string) ([]*meta.Branch, error) {
  // ...
}
```
could be an option. Up to whatever is easiest/most ergonomic for you.

cmd/av/stack_submit.go
```diff
@@ -0,0 +1,82 @@
...
+				actions.CreatePullRequestOpts{
+					BranchName: branchName,
+					Draft:      config.Av.PullRequest.Draft,
+				},
+			)
+			if err != nil {
+				return err
+			}
+			// make sure the base branch of the PR is up to date if it already exists
+			if !result.Created {
```
travigd, line 68: Can we make this only happen if `BaseRefName` is wrong? Also, would be nice to make sure we add `avbeta-stackedprs` label here if it's not set (that should probably be a followup PR though).

internal/actions/pr.go
```diff
@@ -19,6 +19,8 @@ import (
 )
 
 type CreatePullRequestOpts struct {
+	BranchName string
+
```
travigd, line 23: nit: remove unnecessary newline

travigd (approved):
LGTM

image

PR #52, opened 2022-08-03
chore: Setup golangci-lint

travigd:


internal/actions/pr.go
```diff
@@ -277,9 +280,10 @@ func UpdatePullRequestState(ctx context.Context, repo *git.Repo, client *gh.Clie
 	var currentPull *gh.PullRequest
 	// The current open pull request (if any)
 	var openPull *gh.PullRequest
-	for _, pull := range page.PullRequests {
+	for i := range page.PullRequests {
```
jainankit, line 283: was this a bug?

jainankit (approved):

internal/actions/pr.go
travigd: Think so.

It's a pretty common footgun in Go. The loop variable's value changes, not its address/reference. https://go.dev/play/p/BhL5TSaZppX

The most common version of this is when you launch a goroutine in a loop:
```go
for _, val := range vals {
  go func() {
    // do something with val
  }()
}
```

In this case, using `val` is a race condition but it's likely that most goroutines will end up with the *last* value of `val` (since the goroutine captures the loop variable).

travigd: FYI the same thing can happen in Python, though it's a bit more contrived.

```python3
addfns = []
for i in range(10):
    addfns.append(lambda x: x + i)

addfns[9](10)  # 19
addfns[0](10)  # 19
```

PR #54, opened 2022-08-05
feat(stack branch): Support renaming branches

travigd:
This PR is a bit large because golangci-lint just released a new version (last night I think) that breaks our CI for some reason. I'm not entirely sure, but Go 1.19 introduced a small change to how doc comments are formatted when using `gofmt` and `golangci-lint` was complaining (in a rather opaque way) because it... is apparently trying to use that version of `gofmt` even though we're on Go 1.18? I couldn't really reproduce it with `golangci-lint`, but this change just runs the code from source directly. (Adding these kinds of things to `tools.go` is a pretty common practice and doesn't increase the size of our build executable or anything)

If you want to look at just the first commit, that might be easier since it contains all the actual code changes.


cmd/av/stack_branch.go
```diff
@@ -13,6 +14,8 @@ var stackBranchFlags struct {
 	// The parent branch to base the new branch off.
 	// By default, this is the current branch.
 	Parent string
+	// If true, move/rename the current branch.
+	Move bool
```
jainankit, line 18: I'm not sure this is self explanatory. If we are doing renaming, we should just call it `Rename`

cmd/av/stack_branch.go
```diff
@@ -104,3 +111,63 @@ var stackBranchCmd = &cobra.Command{
...
+	currentMeta, _ := meta.ReadBranch(repo, oldBranch)
+	currentMeta.Name = newBranch
+	if err := meta.DeleteBranch(repo, oldBranch); err != nil {
+		return err
+	}
+	if err := meta.WriteBranch(repo, currentMeta); err != nil {
+		return err
+	}
+
+	if !currentMeta.Parent.Trunk {
```
jainankit, line 140: can you add comments describing this part. If I understand this correctly, we are setting the parent to point to the new branch as the child.

jainankit (approved):

PR #56, opened 2022-08-11
feat(stack sync): auto rebase branches after merge

alec-stew-art:
# What changed
- when a stack is partially merged, a stack sync call will now `auto-rebase` onto the trunk commit `tc`
  - if `tc` is the HEAD of trunk, then we simply reparent to trunk
  - if `tc` is not the HEAD of trunk, then we create a new branch with its HEAD at `tc` and reparent to that branch so that stacks continue to work as normal
- added new client pull request functions that access the commit `oid` for both the standard `MERGE` case, as well as the `FAST FORWARD` case.
- added new `GetClient` function that lazy inits the GitHub client with `sync.Once` - switched all functions to use this method but left the old `NewClient` method present if we need it for some reason.

# Testing
Ran the following as a test:
- locally created a stack with three branches stacked on each other: Screen Shot 2022-08-17 at 4 25 58 PM
- ran `stack submit` to create PRs for the full stack: Screen Shot 2022-08-17 at 4 23 36 PM
- ran `/aviator stack merge` on the second PR in the stack leaving the 3rd PR needing to be reparented: Screen Shot 2022-08-17 at 4 28 58 PM
- locally ran `stack sync` and saw the 3rd PR in the stack get auto-rebased: Screen Shot 2022-08-17 at 4 34 52 PM
- this also updated the local state: Screen Shot 2022-08-17 at 4 39 53 PM
- ran `git push --force-with-lease origin mq-test-619133-3` manually because this wasn't yet part of the auto-rebase process (although it now has been added as a step in the code)
- ran `stack submit` and correctly see the PR updated to be on top of `main`: Screen Shot 2022-08-17 at 4 53 49 PM

### DIFF
--- a/cmd/av/stack_sync.go
+++ b/cmd/av/stack_sync.go
@@ -354,6 +354,14 @@ base branch.
 				if err != nil {
 					return err
 				}
+				// force push the updated branch
+				if err := actions.Push(repo, actions.PushOpts{
+					Force:                 actions.ForceWithLease,
+					SkipIfUpstreamNotSet:  true,
+					SkipIfUpstreamMatches: true,
+				}); err != nil {
+					return err
+				}
 				continue loop
 			}
 
--- a/cmd/av/stack_sync.go
+++ b/cmd/av/stack_sync.go
@@ -334,7 +334,7 @@ base branch.
 				}
 				_, _ = fmt.Fprint(os.Stderr,
 					"  - pull request ", colors.UserInput("#", parentMeta.PullRequest.Number),
-					" was merged, this branch should be synced on top of the merge commit:\n",
+					" was merged, syncing branch on top of the merge commit...\n",
 				)
 				// TODO:
 				//     We should do this automatically for the user
@@ -344,7 +344,19 @@ base branch.
 					"        av stack sync --parent ", defaultBranch, "\n",
 				)
 				continue
+			} else if ok && parentMeta.PullRequest != nil && parentMeta.PullRequest.State == githubv4.PullRequestStateClosed {
+				// pull requests are closed on fast-forward so look for a closing commit
+				defaultBranch, err := repo.DefaultBranch()
+				if err != nil {
+					return errors.Wrap(err, "failed to determine default branch")
+				}
+				_, _ = fmt.Fprint(os.Stderr,
+					"  - pull request ", colors.UserInput("#", parentMeta.PullRequest.Number),
+					" was closed, checking for closing commit...\n",
+				)
+				continue
 			}
+
 			_, _ = fmt.Fprint(os.Stderr,
 				"  - syncing ", colors.UserInput(currentBranch),
 				" on top of ", colors.UserInput(currentMeta.Parent.Name), "... ",
--- a/cmd/av/stack_sync.go
+++ b/cmd/av/stack_sync.go
@@ -284,7 +284,7 @@ base branch.
 			}
 
 			// if we have found a related commit in trunk for the PR then skip syncing
-			if currentMeta.TrunkCommit != nil && len(currentMeta.Children) > 0 {
+			if currentMeta.MergeCommit != "" && len(currentMeta.Children) > 0 {
 				_, _ = fmt.Fprint(os.Stderr,
 					"  - pull request ", colors.UserInput("#", currentMeta.PullRequest.Number),
 					" for branch ", colors.UserInput(currentBranch),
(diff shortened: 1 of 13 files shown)


internal/gh/pullrequest.go
```diff
@@ -211,3 +211,79 @@ func (c *Client) RepoPullRequests(ctx context.Context, opts RepoPullRequestOpts)
...
+	} `graphql:"... on Commit"`
+}
+
+type TimelineItemClosedEvent struct {
+	ClosedEvent struct {
+		Closer *TimelineItemCloserCommit
+	} `graphql:"... on ClosedEvent"`
+}
+
+func (c *Client) PullRequestFastForwardCommit(ctx context.Context, opts PullRequestByNumberInput) (*string, error) {
```
alec-stew-art, line 260: @travigd, testing this locally in the GraphQL explorer is not showing the query working like when you showed it to me. I created a test repo and a PR that was merged with fast forward mode but it seems like for me the `closer` on the closing event is always `null` - I have looked at some alternatives, but we will want to discuss this.

internal/gh/client.go
```diff
@@ -32,6 +37,14 @@ func NewClient(token string) (*Client, error) {
 	return &Client{httpClient, githubv4.NewClient(httpClient)}, nil
 }
 
+func GetClient(token string) (*Client, error) {
```
alec-stew-art, line 40: @travigd I think this way of lazy initializing the GH client is a better pattern, but let me know if I missed a super obvious reason why this wouldn't work.

internal/meta/branch.go
```diff
@@ -27,12 +27,24 @@ type Branch struct {
...
+	// The commit onto the trunk branch, if any
+	// This will be the merge commit if the PR was merged
+	TrunkCommit *string `json:"trunkCommit,omitempty"`
 }
 
 func (b *Branch) IsStackRoot() bool {
 	return b.Parent.Trunk
 }
 
+func GetStackRoot(repo *git.Repo, name string) Branch {
```
travigd, line 40: I think it would probably be better to just do `GetStackRoot(branches map[string]Branch, name string) Branch`. Every iteration of this would require a `git` command otherwise which builds up to noticeable latency.

Honestly it probably makes sense to change the abstraction we use here at some point so that we can use some kind of "transactionality". I wrote a big comment in `stack_branch.go` a little while ago:
```go
	// TODO:
	// 		It would be really nice to have some kind of atomicity and/or
	//		transactionality with the branch metadatas. Maybe instead of just
	// 		storing each metadata entry as a blob ref, we could store a single
	// 		tree ref that points to each of the branch blobs (since it's
	//	 	possible to update the tree refs atomically w/ Git's CAS).
	// 		It's possible that we run the operations above but then this fails
	//		and we're in a weird state.
```

internal/gh/pullrequest.go
```diff
@@ -211,3 +211,79 @@ func (c *Client) RepoPullRequests(ctx context.Context, opts RepoPullRequestOpts)
...
+	Owner  string
+	Repo   string
+	Number int64
+}
+
+type Commit struct {
+	Oid string
+}
+
+func (c *Client) PullRequestMergeCommit(ctx context.Context, opts PullRequestByNumberInput) (*string, error) {
```
travigd, line 225: It probably makes sense to just use the existing `Client.PullRequest` method for this and just unconditionally include this information.

internal/gh/pullrequest.go
travigd: Yeah, let's chat. Will also say that we should probably try to do everything in a single API request rather than having two different ones.

cmd/av/stack_sync.go
```diff
@@ -327,24 +320,55 @@ base branch.
 			})
 
 			parentMeta, ok := branches[parentState.Name]
-			if ok && parentMeta.PullRequest != nil && parentMeta.PullRequest.State == githubv4.PullRequestStateMerged {
+			if ok && parentMeta.TrunkCommit != nil {
 				defaultBranch, err := repo.DefaultBranch()
```
travigd, line 324: We should actually figure out the name of the trunk branch when we start the sync and not assume it's the default branch name here. But probably outside the scope of this PR.

travigd (commented):
I think reparenting is maybe not quite the right approach here (at least semantically). I think a "vanilla rebase" is going to be a simpler thing to do (at least conceptually). Happy to elaborate more if that's not totally clear.

internal/gh/pullrequest.go
```diff
@@ -24,6 +24,20 @@ type PullRequest struct {
 	Permalink   string
 	State       githubv4.PullRequestState
 	Title       string
+	MergeCommit struct {
```
travigd, line 27: Can you change this from `MergeCommit struct {...}` to `` PRIVATE_MergeCommit struct {..} `graphql:"mergeCommit"` ``? Just to indicate to calling code that we really shouldn't be accessing it directly.

Then renaming the function to just `MergeCommit` is good.

travigd: Similar for `TimeLineItems` -- also lets do `PRIVATE_TimelineItems` there (no capital `L` to match GitHub here)

travigd (approved):
Looks good! Few minor comments. I assume you've tested this.

internal/actions/reparent.go
```diff
@@ -114,15 +115,15 @@ func ReparentContinue(repo *git.Repo, opts ReparentOpts) (*ReparentResult, error
...
+		if err := ReparentWriteMetaData(repo, opts); err != nil {
 			return nil, err
 		}
 		return &ReparentResult{Success: true}, nil
 	}
 	return handleReparentRebaseOutput(repo, opts, output)
 }
 
-func reparentWriteMetadata(repo *git.Repo, opts ReparentOpts) error {
+func ReparentWriteMetaData(repo *git.Repo, opts ReparentOpts) error {
```
travigd, line 126: nit: `Metadata` rather than `MetaData`

cmd/av/stack_sync.go
```diff
@@ -327,24 +319,58 @@ base branch.
...
+				_, err = repo.Rebase(git.RebaseOpts{
+					Onto:     parentMeta.MergeCommit,
+					Upstream: parentMeta.Name,
+					Branch:   currentBranch,
+				})
+				if err != nil {
+					return err
+				}
+				// now that we have rebased onto trunk - update current branch
+				err = actions.ReparentWriteMetaData(repo, actions.ReparentOpts{
```
travigd, line 350: Oh I missed this. I don't think this is the right function to be calling here. `reparentWriteMetadata` was a private function for a reason. I don't think we should think of this as a reparent operation at all internally. Can we just use the normal `meta.WriteBranch` here?

PR #57, opened 2022-08-25
feat: Add stack metadata to PR on creation

travigd:
I think this will help us be more robust against weird things happen because we're trying to figure out the users intention based on how the PRs are arranged on GitHub.

This also will mean that we don't have to use the `avbeta-stackedprs` label since we can just read the metadata from the PR comment. I think this will be necessary to avoid accidentally injecting stacked PRs functionality into existing dev workflows.

This is what it looks like:
image

### DIFF
--- a/internal/actions/pr.go
+++ b/internal/actions/pr.go
@@ -147,6 +147,9 @@ func CreatePullRequest(ctx context.Context, repo *git.Repo, client *gh.Client, o
 	}
 
 	trunkBranch, err := meta.Trunk(repo, opts.BranchName)
+	if err != nil {
+		return nil, errors.WrapIf(err, "failed to determine trunk branch")
+	}
 	prMeta := WritePRMetadata(PRMetadata{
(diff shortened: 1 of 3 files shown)


internal/actions/pr.go
```diff
@@ -179,6 +185,7 @@ func CreatePullRequest(ctx context.Context, repo *git.Repo, client *gh.Client, o
 	if didCreatePR {
 		action = "created"
 	} else {
+		// TODO: need to update the PR metadata possibly here
```
doratzeng, line 188: what happens right now if we don't update the metadata here?

travigd: Nothing right now since we don't use this on the server side. My plan is to finish this TODO then change the Aviator server side so that it's reading this information instead of trying to implicitly determine the shape of the stack based on how the PRs are arranged.

doratzeng (approved):

PR #58, opened 2022-08-25
feat(stack sync): Implement av stack sync --trunk

travigd:
Tested by running `av stack sync --trunk`:
* From the root of the stack
* From a child branch of the stack
* When root is up-to-date with trunk
* When root is not up-to-date with trunk
* When root has conflicts with trunk


doratzeng (approved):
please also update documentation!

PR #59, opened 2022-09-12
feat: Include parent pull in stack metadata, update metadata on stack submit

travigd:
This PR adds `parentPull` to stack metadata. It also will update the metadata whenever `CreatePullRequest` is called (_i.e._, when doing `av pr create` or `av stack submit`).

We'll probably need to detect if things are bad™ on the server side of things and recommend running `av stack submit`, but at least this should make it so that the logic for constructing stacks is more explicit and intentional and consistent and less likely to do unexpected things.

### DIFF
--- a/internal/actions/pr.go
+++ b/internal/actions/pr.go
@@ -48,6 +48,10 @@ type CreatePullRequestResult struct {
 // CreatePullRequest creates a pull request on GitHub for the current branch, if
 // one doesn't already exist.
 func CreatePullRequest(ctx context.Context, repo *git.Repo, client *gh.Client, opts CreatePullRequestOpts) (*CreatePullRequestResult, error) {
+	if opts.BranchName == "" {
+		logrus.Panicf("internal invariant error: CreatePullRequest called with empty branch name")
+	}
+
 	repoMeta, err := meta.ReadRepository(repo)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


internal/actions/pr.go
```diff
@@ -69,7 +73,7 @@ func CreatePullRequest(ctx context.Context, repo *git.Repo, client *gh.Client, o
 		//       break if the upstream name is not the same name as the local
 		upstream, err := repo.RevParse(&git.RevParse{
 			SymbolicFullName: true,
-			Rev:              "HEAD@{u}",
+			Rev:              fmt.Sprintf("%s@{u}", opts.BranchName),
```
travigd, line 76: this is an unrelated bug: we would always push whatever branch was currently checked out rather than the target branch.

Bad behavior (that's now fixed):
```
$ av stack submit
Creating pull request for branch 2022-09-12-one:
  - pushing branch to GitHub (origin/2022-09-12-three)
...
```
(note that the `pushing branch to` bit is wrong)

internal/actions/pr.go
```diff
@@ -190,19 +202,16 @@ func CreatePullRequest(ctx context.Context, repo *git.Repo, client *gh.Client, o
...
-		action = "fetched existing"
+		action = "sychronized"
 	}
 	_, _ = fmt.Fprint(os.Stderr,
-		"  - ", action, " pull request for branch ", colors.UserInput(opts.BranchName),
-		" (into branch ", colors.UserInput(prBaseBranch), "): ",
-		colors.UserInput(pull.Permalink),
-		"\n",
+		"  - ", action, " pull request ",
+		colors.UserInput(pull.Permalink), "\n",
```
travigd, line 209: tweaked this just to be less visually noisy.
Screen Shot 2022-09-12 at 11 34 52 AM

alec-stew-art (approved):

PR #61, opened 2022-09-27
fix: Update base branch during av stack sync

travigd:
image

I don't like the diverging-convering logic of this where we have the `syncBranchUpdateNewTrunk` being called from `syncBranchRebase` **and** `syncBranchContinue`. Might try to think about how to do it better in the future, but this is probably fine for now.


alec-stew-art (approved):
lgtm

PR #62, opened 2022-09-30
fix: Construct correct list of branches to be synced during stack sync

travigd:
This fixes how we store the list of branches being synced during the current `av stack sync`. This bug would only manifest if the stack sync was being run from a branch that is not at the end of the stack and the stack sync was interrupted on a branch that comes after the originally checked out branch in the stack.

Confirmed that the new test failed with the code changes and passes now.


ohcnivek (approved):
nice nice

PR #63, opened 2022-10-04
fix: Stack sync rebase properly after parent is merged

travigd:

### DIFF
--- a/internal/meta/branch.go
+++ b/internal/meta/branch.go
@@ -60,10 +60,14 @@ func (b *Branch) UnmarshalJSON(bytes []byte) error {
 		return err
 	}
 
-	// Everything except name (since that is set externally) and parent is set
-	// on the actual Branch object we're unmarshalling into
-	b.Children = d.Children
-	b.PullRequest = d.PullRequest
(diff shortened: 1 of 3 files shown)


internal/meta/branch.go
```diff
@@ -60,21 +60,21 @@ func (b *Branch) UnmarshalJSON(bytes []byte) error {
...
-	// on the actual Branch object we're unmarshalling into
-	b.Children = d.Children
-	b.PullRequest = d.PullRequest
+	// Copy over all the data that we unmarshalled into BranchAlias. This is
+	// everything except since Parent which we'll handle next. We need to take
+	// special care to copy name since it won't be defined on BranchAlias (since
+	// Name is not serialized to JSON). Instead we expect that the struct is
+	// always initialized with the name defined, so we have to copy it over here.
+	// (Doing just "*b = ..." will result in us erasing the name.)
+	d.BranchAlias.Name = b.Name
```
travigd, line 69: Actual bug here was that we weren't copying `MergeCommit` from `BranchAlias`. I changed it to just copy everything and then update the `Parent` field so that this kind of bug doesn't happen again.

ohcnivek (approved):

PR #71, opened 2023-01-11
fix: Don't un-draft PRs during re-parent

travigd:
Fixing this as reported by a customer.

Before this, we would actually un-draft (aka _mark ready for review_) PRs that were in a draft state if we were re-parenting the PR.

Incidentally, while fixing that bug, I realized that we actually were never transitioning PR's to a draft state at all when they should have been.

With this PR, I verified that now:
* PRs that should be switched to draft during re-parent (to avoid adding unnecessary codeowners) actually are
* PRs that are in a draft state when the re-parent runs continue to stay draft PRs

### DIFF
--- a/cmd/av/stack_sync.go
+++ b/cmd/av/stack_sync.go
@@ -149,7 +149,6 @@ base branch.
 			return errors.New("refusing to sync: there are unstaged changes in the working tree (use `git add` to stage changes)")
 		}
 
-		var currentBranch string
 		if stackSyncFlags.Continue {
 			if state.CurrentBranch == "" {
 				return errors.New("no sync in progress")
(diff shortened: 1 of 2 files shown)


cmd/av/stack_sync.go
```diff
@@ -149,7 +149,6 @@ base branch.
 			return errors.New("refusing to sync: there are unstaged changes in the working tree (use `git add` to stage changes)")
 		}
 
-		var currentBranch string
```
travigd, line 152: also found a bug where we weren't always setting `currentBranch` so to make the code a little simpler I just removed `currentBranch` and make sure we always set `state.CurrentBranch` and only refer to that.

alec-stew-art (approved):

PR #73, opened 2023-02-06
fix: Parse EDITOR with shell syntax

travigd:
Fixing a bug reported by a customer.

Now we parse the result of `EDITOR`/`GIT_EDITOR` with shell syntax.

This is what [GitHub's `hub` CLI does](https://github.com/github/hub/issues/721) as well as what Git itself does:
```
$ git help var
...
       GIT_EDITOR
           Text editor for use by Git commands. The value is meant to be interpreted by the shell when it is used. Examples: ~/bin/vi, $SOME_ENVIRONMENT_VARIABLE,
           "C:\Program Files\Vim\gvim.exe" --nofork. The order of preference is the $GIT_EDITOR environment variable, then core.editor configuration, then $VISUAL,
           then $EDITOR, and then the default chosen at compile time, which is usually vi.
...
```

This PR adds some tests and I'm actually submitting the PR with `EDITOR="code --wait" av pr create --draft`. Notably, it doesn't support the `$SOME_ENVIRONMENT_VARIABLE` syntax for the editor, but that's probably fine (at least for now).

### DIFF
--- a/internal/editor/editor.go
+++ b/internal/editor/editor.go
@@ -3,7 +3,9 @@ package editor
 import (
 	"bufio"
 	"bytes"
+	"emperror.dev/errors"
 	"github.com/aviator-co/av/internal/git"
+	"github.com/kballard/go-shellquote"
 	"github.com/sirupsen/logrus"
 	"os"
 	"os/exec"
@@ -23,12 +25,22 @@ type Config struct {
 	Command string
 }
 
+// CommandNoOp is a special command that indicates that no editor should be
+// launched and the text should be returned as-is.
+// This behavior is copied from git's GIT_EDITOR.
+// https://github.com/git/git/blob/5699ec1b0aec51b9e9ba5a2785f65970c5a95d84/editor.c#L57
+const CommandNoOp = ":"
+
 func Launch(repo *git.Repo, config Config) (string, error) {
 	switch {
 	case config.Command == "":
 		config.Command = DefaultCommand(repo)
 	case config.TmpFilePattern == "":
 		config.TmpFilePattern = "av-message-*"
+	}
+
+	if config.Command == CommandNoOp {
+		return config.Text, nil
 	}
 
 	tmp, err := os.CreateTemp("", config.TmpFilePattern)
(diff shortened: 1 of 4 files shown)


internal/editor/editor.go
```diff
@@ -47,8 +59,17 @@ func Launch(repo *git.Repo, config Config) (string, error) {
...
+	// We interpret the command with shell syntax to allow users to specify
+	// both flags and use editor executables with spaces.
+	// e.g., EDITOR="'/path/with spaces/editor'" or
+	// EDITOR="code --wait" work.
+	args, err := shellquote.Split(config.Command)
+	if err != nil {
+		return "", errors.Wrapf(err, "invalid editor command: %q", config.Command)
+	}
+	args = append(args, tmp.Name())
+	cmd := exec.Command(args[0], args[1:]...)
```
jainankit, line 72: you are not using `tmp.Name()` anymore, is that intentional?

internal/editor/editor.go
```diff
@@ -47,8 +59,17 @@ func Launch(repo *git.Repo, config Config) (string, error) {
...
 	}
 
-	// launch the editor as a subprocess
-	cmd := exec.Command(config.Command, tmp.Name())
+	// Launch the editor as a subprocess.
+	// We interpret the command with shell syntax to allow users to specify
+	// both flags and use editor executables with spaces.
+	// e.g., EDITOR="'/path/with spaces/editor'" or
+	// EDITOR="code --wait" work.
+	args, err := shellquote.Split(config.Command)
```
jainankit, line 67: So if I understand this correctly, you are using a different interpreter for executing the command? Is that why code --wait didn't work for Nick?

travigd: `shellquote.Split` just converts a string into arguments. So an input of `"/foo bar/spam" --eggs` gets converted to `[]string{"/foo bar/spam", "--eggs"}`. There's no "interpreting" happening here, it just applies shell string quoting rules to convert the string into a list of arguments (which ultimately forms the `argv` of the command that's executed).

internal/editor/editor.go
travigd: Yeah, now it's inside `args` (via the `args = append(args, ...)` the line above). Go isn't quite as ergonomic as Python about when you can splat things (I think to get more control over when it does allocations and whatnot).

```go
// The function is defined as
func Command(name string, arg ...string) *Cmd

// Illegal because the first argument can't be given as part of the splat
cmd := exec.Command(args..., "foo")

// Illegal because you can't add things after a splat
cmd := exec.Command(args[0], args[1:]..., "foo")

// Legal because there's no splat and Go can coalesce the varags into a single array
cmd := exec.Command("foo", "bar", "spam", "eggs")
// equivalent to
cmd := exec.Command("foo", []string{"bar", "spam", "eggs"}...)
```

internal/editor/editor.go
jainankit: yeh I think i inferred that, the question was rather why that was causing `Command` to fail, I'm guessing this is how it works. The first param itself need to be the actual command (without params).

jainankit (approved):

PR #80, opened 2023-04-13
feat: Add commit split command

draftcode:

### DIFF
--- a/cmd/av/commit_split.go
+++ b/cmd/av/commit_split.go
@@ -1,0 +1,156 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"emperror.dev/errors"
+	"github.com/aviator-co/av/internal/git"
+	"github.com/aviator-co/av/internal/utils/colors"
+	"github.com/spf13/cobra"
+)
+
+var commitSplitCmd = &cobra.Command{
+	Use:          "split",
+	Short:        "split a commit into multiple commits",
+	SilenceUsage: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		repo, err := getRepo()
+		if err != nil {
+			return err
+		}
+		if clean, err := repo.CheckCleanWorkdir(); err != nil {
+			return err
+		} else if !clean {
+			_, _ = fmt.Fprint(
+				os.Stderr,
+				colors.Failure("The working directory is not clean, please stash or commit them before running split command."),
+			)
+			return errors.New("the working directory is not clean")
+		}
+
+		// Ignore errors to support a detached HEAD.
+		currentBranchName, _ := repo.CurrentBranchName()
+		currentCommitOID, err := repo.RevParse(&git.RevParse{Rev: "HEAD"})
+		if err != nil {
+			return errors.Errorf("cannot get the current commit object: %v", err)
+		}
+
+		// From here, we use detached HEAD, so that even if something goes wrong or user
+		// aborts the operation in the middle, the original branch is intact.
+		if err := splitCommit(repo, currentBranchName, currentCommitOID); err != nil {
+			commitSplitAbortMessage(currentBranchName, currentCommitOID)
+			return err
+		}
+		return nil
+	},
+}
+
+func splitCommit(repo *git.Repo, currentBranchName, currentCommitOID string) error {
+	if _, err := repo.Git("switch", "--detach", currentCommitOID); err != nil {
+		return err
+	}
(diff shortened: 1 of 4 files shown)


cmd/av/commit_split.go
```diff
@@ -0,0 +1,142 @@
...
+	SilenceUsage: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		repo, err := getRepo()
+		if err != nil {
+			return err
+		}
+		if clean, err := repo.CheckCleanWorkdir(); err != nil {
+			return err
+		} else if !clean {
+			return errors.New("the workdir is not clean")
```
jainankit, line 25: if this is a user facing message, you should explain it better
```suggestion
			return errors.New("The working directory is not clean, please stash or commit them before running split command")
```

jainankit: Also what's the difference between `errors.Errorf` and `errors.New`?

cmd/av/commit_split.go
```diff
@@ -0,0 +1,142 @@
...
+			return errors.Errorf("cannot update the branch %s to the new commit %s: %v", currentBranchName, newCommitOID, err)
+		}
+		// At this point, the HEAD is still a detached HEAD. Check out the branch.
+		// repo.CheckoutBranch errors out if the repository is at the detached head.
+		// We have to run git checkout in other ways.
+		if _, err := repo.Git("switch", currentBranchName); err != nil {
+			return errors.Errorf("cannot switch to the original branch: %v", err)
+		}
+
+		// TODO: We should rebase the stacks after split.
```
jainankit, line 109: can you create an issue for this in GH?

internal/git/git.go
```diff
@@ -156,6 +156,29 @@ func (r *Repo) CurrentBranchName() (string, error) {
 	return branch, nil
 }
 
+// CheckCleanWorkdir returns if the workdir is clean.
+func (r *Repo) CheckCleanWorkdir() (bool, error) {
+	out, err := r.Git("status", "--porcelain")
+	if err != nil {
+		return false, errors.Errorf("failed to check if the workdir is clean: %v", err)
```
jainankit, line 163:
```suggestion
		return false, errors.Errorf("failed to check if the working directory is clean: %v", err)
```

jainankit (approved):

cmd/av/commit_split.go
travigd: See [`errors.Errorf`](https://pkg.go.dev/emperror.dev/errors#Errorf) and [`errors.New`](https://pkg.go.dev/emperror.dev/errors#New).

The tl;dr: is that `Errorf` allows using a format string while `New` only takes a single string argument.

cmd/av/commit_split.go
```diff
@@ -0,0 +1,142 @@
...
+			colors.UserInput(commitOID), "\n",
+			"\n",
+			"In order to get back to ", colors.UserInput(commitOID), ",\n",
+			"you need to run ", colors.CliCmd("git switch --discard-changes --detach", commitOID), "\n",
+			colors.Failure("===================================================="), "\n",
+		)
+		return
+	}
+	_, _ = fmt.Fprint(
+		os.Stderr,
```
travigd, line 132: I think `RECOVERY INSTRUCTION` is a bit too severe of a message here. Makes people think they're in a dangerous state and they could lose their work.

Maybe something like
```
====================================================
DETACHED HEAD

The commit split command aborted.
Your original branch {branch} was not modified.
Your Git repository is now in a detached HEAD state.
To revert your changes and switch to your original (unmodified) branch, run:

    git switch --discard-changes {branch}
====================================================

File: cmd/av/commit_split.go (hunk 2 above)
Comment (travigd) [reply cmd/av/commit_split.go:109]: At the very least, can you print out a message to the user that says
```
Run `av stack sync` to sync your stack if necessary.
```

travigd (approved):

cmd/av/commit_split.go
draftcode: Done.

internal/git/git.go
draftcode: Done.

cmd/av/commit_split.go
draftcode: Done.

cmd/av/commit_split.go
draftcode: Done.

PR #81, opened 2023-04-13
feat: Use the remote tracking branch for finding commits

draftcode:
When a branch parent is trunk, use the remote tracking branch instead of
using the locally checked out trunk branch. This makes more accurate
commit list, and in case where the user intentionally removes local
trunk branch to avoid unintentional commit on master/main, av pr create
starts working.

### DIFF
--- a/internal/actions/pr.go
+++ b/internal/actions/pr.go
@@ -5,13 +5,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/aviator-co/av/internal/editor"
-	"github.com/aviator-co/av/internal/utils/stringutils"
-	"github.com/aviator-co/av/internal/utils/templateutils"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


internal/actions/pr.go
```diff
@@ -137,6 +138,7 @@ func CreatePullRequest(ctx context.Context, repo *git.Repo, client *gh.Client, o
 	// figure this out based on whether or not we're on a stacked branch
 	branchMeta, _ := meta.ReadBranch(repo, opts.BranchName)
 	prBaseBranch := branchMeta.Parent.Name
+	localPRBaseBranch := prBaseBranch
```
travigd, line 141: I think something like `prCompareRef` might be a better name for this.

travigd (approved):

internal/actions/pr.go
draftcode: Done.

PR #83, opened 2023-04-14
feat: Show the commit message body

draftcode:
Currently we show the commit message subjects (the first line of the
message) as the comment in the PR message editor. This shows the message
body as well. This is useful when there are multiple commits or when
there is a PR template (.github/PULL_REQUEST_TEMPLATE.md) exists.

### DIFF
--- a/internal/actions/pr.go
+++ b/internal/actions/pr.go
@@ -270,7 +270,9 @@ type prBodyTemplateData struct {
 	Commits []git.CommitInfo
 }
 
-var prBodyTemplate = template.Must(template.New("prBody").Parse(`%% Creating pull request for branch '{{ .Branch }}'
+var templateFuncs = template.FuncMap{"trimSpace": strings.TrimSpace}
+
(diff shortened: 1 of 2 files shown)


travigd (commented):
Is there context for why you're adding this change? I'd imagine the commit titles would be enough to see what you're about to open a PR for. I worry that including the commit messages will make it harder to read.

draftcode:
I usually make one commit that 1st line as a title and the rest is the detail. If there is no PR template, av-cli takes them as the title and description.

The problem is when there is a PR template. Then the description is populated with the PR template and I cannot see the commit's detail part.

I think this change wouldn't affect multiple commits readability much because I guess if people put multiple commits, they would only put titles. This change won't make any difference if a commit message is only one line.

If a person puts more than one line on a commit message and they want to have them reviewed in one PR, this change will have lengthy pre-populated text (as a comment). First of all, I feel like if each commit requires such a lengthy explanation, it's better to be an independent PR, not one PR. But even if they want them to be in one PR, it's helpful to explain the overall PR description to reference individual commits (if each commit has such huge changes to that requires more than one line description).

travigd:
Fair enough!

travigd (approved):

PR #85, opened 2023-04-17
feat: Add av stack tidy

draftcode:
This command is used to update the av-cli's internal branch metadata
based on the Git branches to some extent. The expected usage is that
after a branch is merged or deleted, run this command to synchronize the
internal metadata.

Probably we can do more magic and hide the details on the branch
metadata furthermore, and automatically tidies up in av stack sync. That
can be done in the future. Right now, unable to delete a branch is
critical, and this should act as a stop-gap solution.

### DIFF
--- a/cmd/av/stack_tidy.go
+++ b/cmd/av/stack_tidy.go
@@ -1,0 +1,98 @@
+package main
+
+import (
+	"strings"
+
+	"github.com/aviator-co/av/internal/git"
+	"github.com/aviator-co/av/internal/meta"
+	"github.com/spf13/cobra"
+)
+
+var stackTidyCmd = &cobra.Command{
+	Use:          "tidy",
+	Short:        "tidy up the branch metadata",
+	Long: strings.TrimSpace(`
+Tidy up the branch metadata by removing the deleted / merged branches.
+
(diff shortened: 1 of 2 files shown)


cmd/av/stack_tidy.go
```diff
@@ -0,0 +1,98 @@
...
+		repo, err := getRepo()
+		if err != nil {
+			return err
+		}
+		origBranches, err := meta.ReadAllBranches(repo)
+		if err != nil {
+			return err
+		}
+		branches := make(map[string]*meta.Branch)
+		for name, br := range origBranches {
```
travigd, line 31: I don't think this is strictly necessary. We could just do

```
for name := range branches {
    branches[name].Parent = newParent
}
```

travigd: If you feel that you'd rather do it this way, that's fine too.

travigd (approved):

cmd/av/stack_tidy.go
draftcode: Nope. You cannot. See https://stackoverflow.com/questions/42605337/cannot-assign-to-struct-field-in-a-map

PR #87, opened 2023-04-19
feat: Support detached HEAD in stack tree

draftcode:

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -144,6 +144,19 @@ func (r *Repo) GitStdin(args []string, stdin io.Reader) (string, error) {
 	return strings.TrimSpace(string(out)), nil
 }
 
+// DetachedHead returns true if the repository is in the detached HEAD.
+func (r *Repo) DetachedHead() (bool, error) {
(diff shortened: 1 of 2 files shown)


internal/git/git.go
```diff
@@ -144,6 +144,15 @@ func (r *Repo) GitStdin(args []string, stdin io.Reader) (string, error) {
 	return strings.TrimSpace(string(out)), nil
 }
 
+// DetachedHead returns true if the repository is in the detached HEAD.
+func (r *Repo) DetachedHead() (bool, error) {
+	name, err := r.RevParse(&RevParse{Rev: "HEAD", SymbolicFullName: true})
```
travigd, line 149: Can we do `git symbolic-ref -q HEAD` instead? Seems more explicit. https://stackoverflow.com/questions/52221558/programmatically-check-if-head-is-detached

travigd (approved):

internal/git/git.go
draftcode: Done.

PR #89, opened 2023-04-21
feat: Add hidden branch-meta commands

draftcode:
This adds hidden branch-meta commands for the internal metadata access.
Useful for debugging / addressing user issues.

This neglects the parent-child relationship maintenance and just try to
rebuild the relationship from parents for all branches.

### DIFF
--- a/cmd/av/branchmeta.go
+++ b/cmd/av/branchmeta.go
@@ -1,0 +1,147 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"emperror.dev/errors"
+	"github.com/aviator-co/av/internal/git"
+	"github.com/aviator-co/av/internal/meta"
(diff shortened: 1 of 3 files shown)


travigd (approved):
Can we change the `av branch-meta set-parent` command to just `av branch-meta set` and that way we can add other options in the future?

```sh
av branch-meta set <branch name> --parent=<parent branch name>
```

Also, adding `av branch-meta rebuild-children` as a command would be good IMO.

draftcode:
> Can we change the `av branch-meta set-parent` command to just `av branch-meta set` and that way we can add other options in the future?
> 
> ```shell
> av branch-meta set <branch name> --parent=<parent branch name>
> ```
> 
> Also, adding `av branch-meta rebuild-children` as a command would be good IMO.

Done.

PR #90, opened 2023-04-24
feat: allow editing a PR description

draftcode:
Previously, if there's an existing PR, an editor pops open, but the
content is discarded anyway. Now it won't opens an editor in such case.

With the new flag `--edit`, an editor opens for an existing PR. The
content is pre-filled with the existing PR description and updated
contents are reflected to the PR.

### DIFF
--- a/cmd/av/pr.go
+++ b/cmd/av/pr.go
@@ -117,6 +117,6 @@ func init() {
 	)
 	prCreateCmd.Flags().BoolVar(
 		&prCreateFlags.Edit, "edit", false,
-		"open the editor for editing always",
+		"always open an editor to edit the pull request title and description",
 	)
 }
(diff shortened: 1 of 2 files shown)


cmd/av/pr.go
```diff
@@ -113,4 +115,8 @@ func init() {
 		&prCreateFlags.Body, "body", "b", "",
 		"body of the pull request to create (a value of - will read from stdin)",
 	)
+	prCreateCmd.Flags().BoolVar(
+		&prCreateFlags.Edit, "edit", false,
+		"open the editor for editing always",
```
travigd, line 120:
```suggestion
		"always open an editor to edit the pull request title and description",
```

travigd (approved):

PR #91, opened 2023-04-25
feat: [MER-2190] implement av stack diff

doratzeng:


cmd/av/stack_diff.go
```diff
@@ -0,0 +1,62 @@
...
+Generates the diff between the working tree and the current index (i.e., the diff containing all unstaged changes).
+`),
+	SilenceUsage: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		repo, err := getRepo()
+		if err != nil {
+			return err
+		}
+
+		diff, err := repo.Diff(&git.DiffOpts{})
```
travigd, line 26: This needs an argument for `Commit` otherwise we'll only ever generate the diff between the previous commit and the working tree (so there's no reason to run `av stack diff` versus `git diff`).

We need to determine that commit by looking at the current branch, finding the parent (if any -- the branch might be a stack root), and comparing against that.

There's a bit of a wrinkle about what happens if the branch isn't up to date with its parent (i.e., the user needs to run `av stack sync`), but we can deal with that later.

travigd (commented):
The Go looks good! Few notes:

We shouldn't use `Success`/`Failure` to print the added/removed lines. Those are semantic names but adding a line of code is not necessarily a success. Instead we should just define new colors named `Green`/`Red`.

Thats kind of moot though, because I think we can just pass through the colors from `git` directly here. We can add a `Color bool` field to `DiffOpts` and then when we call it, we can do:
```

```
repo.Diff(&git.DiffOpts{
    Color: !color.NoColor,  // color from github.com/fatih/color module
})

cmd/av/stack_diff.go
```diff
@@ -23,40 +25,25 @@ Generates the diff between the working tree and the current index (i.e., the dif
...
-				)
-			}
+		branch, _ := meta.ReadBranch(repo, currentBranchName)
+
+		var opts = &git.DiffOpts{
+			Color: !color.NoColor,
+			Commit: branch.Parent.Name,
+		}
+
+		diff, err := repo.Diff(opts)
```
travigd, line 40: nit: don't create a variable for `opts`, just declare the struct directly here. It's basically just the Go way of specifying keyword arguments.

travigd (approved):
💯

PR #92, opened 2023-04-25
refactor: Create meta.DB interface

travigd:
This is just the first PR in what will probably be a few. This one adds a basic database interface that just stores everything as a big `map` (which is probably fine for our scale of data) and writes to/reads from a JSON file as the persistence mechanism.

Wanted to open this now and get some feedback. In the next PR, I'll probably implement writing to refs while also writing to the JSON database. After that, the annoying change of having to plumb the database through everywhere.

### DIFF
--- a/internal/meta/db.go
+++ b/internal/meta/db.go
@@ -8,6 +8,8 @@ type DB interface {
 // ReadTx is a transaction that can be used to read from the database.
 // It presents a consistent view of the underlying database.
 type ReadTx interface {
+	// Repository returns the repository information.
+	Repository() (Repository, bool)
 	// Branch returns the branch with the given name. If no such branch exists,
 	// the second return value is false.
 	Branch(name string) (Branch, bool)
@@ -28,4 +30,6 @@ type WriteTx interface {
 	Commit() error
 	// SetBranch sets the given branch in the database.
 	SetBranch(branch Branch)
+	// SetRepository sets the repository information in the database.
+	SetRepository(repository Repository)
 }
--- a/internal/meta/db.go
+++ b/internal/meta/db.go
@@ -1,0 +1,26 @@
+package meta
+
+type DB interface {
+	ReadTx
+	// WithTx runs the given function in a transaction. The transaction is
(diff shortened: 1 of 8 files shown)


internal/meta/db.go
```diff
@@ -0,0 +1,26 @@
+package meta
+
+type DB interface {
+	ReadTx
+	// WithTx runs the given function in a transaction. The transaction is
+	// committed when the function returns (unless it has been aborted by
+	// calling WriteTx.Abort).
+	WithTx(func(tx WriteTx)) error
```
draftcode, line 8: `WithTx(func(tx WriteTx) error)) error` might be more convenient as we can use it in RunE directly.

```
RunE: func(...) error {
    db := ...
    return db.WriteTx(func(tx WriteTx) error) error {
        ...
    })
}
```

travigd: That was actually my original strategy but I chose against it to make it more clear that it's not a direct passthrough -- `WriteTx` can returns its own errors. Honestly, in hindsight, I think just doing a
```go
tx := db.WriteTx()
tx.SetBranch(...)
tx.Commit()
```
will be a better API. I chose the function-arg version to make it impossible to forget to commit/abort a transaction, but I think it's fine to just have an API invariant of "you MUST call `Commit` or `Abort` on this transaction object".

draftcode (approved):
Rereviewed for the updated interface.

internal/meta/db.go
```diff
@@ -0,0 +1,35 @@
...
+	AllBranches() map[string]Branch
+}
+
+// WriteTx is a transaction that can be used to modify the database.
+type WriteTx interface {
+	ReadTx
+	// Abort aborts the transaction (no changes will be committed).
+	// The transaction cannot be used after it has been aborted.
+	Abort()
+	// Commit commits the transaction (all changes will be committed).
```
draftcode, line 26: Small ask: Can we add a contract like "Abort can be called even after Commit". Otherwise, we need to cancel the Cleanup.

PR #93, opened 2023-04-25
refactor: internally use the positive flags

draftcode:
While the flags are --no-fetch/--no-push, internally use the positive
flags because it's hard to read.


travigd (approved):
We talked about this externally and I think it's a decent change.

There's a bit of a distinction between how we want to handle this internally/externally (just writing this for posterity):
* For external things (i.e., the public API of the `av` CLI), we optimize for the ideal UX. Generally, we want to fetch/push by default otherwise users will be surprised when their changes after `sync` are not reflected on GitHub.
* For internal things, we want to optimize both for code readability and also for opting-in to "side effects." Generally, not fetching/not pushing is the "safer" default since it doesn't modify external systems (such as the local Git database or GitHub).

Either way, we still have default values are false in typical Go style.

PR #96, opened 2023-04-28
feat: [MER-1493] show up to date info about branches in stack tree

doratzeng:
add the following info about each branch for `av stack tree` (will probably need some UI improvements)
- is it up to date with parent branch
- is it up to date with upstream/remote branch
- if PR link if it exists

Screen Shot 2023-05-01 at 2 12 32 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,4 +9,3 @@ To run the command line:
 ```
 go run ./cmd/av [subcommand/flags...]
 ```
-testing testing
--- a/README.md
+++ b/README.md
@@ -9,3 +9,4 @@ To run the command line:
 ```
 go run ./cmd/av [subcommand/flags...]
 ```
+testing testing
--- a/cmd/av/stack_tree.go
+++ b/cmd/av/stack_tree.go
@@ -69,7 +69,11 @@ func printStackTree(repo *git.Repo, branches map[string]meta.Branch, currentBran
 		return
 	}
 
-	branchInfo := getBranchInfo(repo, branch)
+	branchInfo, err := getBranchInfo(repo, branch)
+	if err != nil {
+		fmt.Printf("<ERROR: cannot get branch info: %v>\n", err)
+		return
+	}
 
 	if currentBranch == branch.Name {
 		_, _ = fmt.Print(
@@ -83,19 +87,19 @@ func printStackTree(repo *git.Repo, branches map[string]meta.Branch, currentBran
 	}
 }
 
-func getBranchInfo(repo *git.Repo, branch meta.Branch) string {
+func getBranchInfo(repo *git.Repo, branch meta.Branch) (string, error) {
 	var branchInfo string
 
 	parentStatus, err := getParentStatus(repo, branch)
 	if err != nil {
 		fmt.Printf("<ERROR: could not get status of parent branch: %v>\n", err)
-		return ""
+		return "", err
 	}
 
 	upstreamStatus, err := getUpstreamStatus(repo, branch)
 	if err != nil {
 		fmt.Printf("<ERROR: could not get status of upstream branch: %v>\n", err)
-		return ""
+		return "", err
 	}
 
 	branchStatus := strings.Trim(fmt.Sprintf("%s, %s", parentStatus, upstreamStatus), ", ")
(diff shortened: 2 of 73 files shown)


cmd/av/stack_tree.go
```diff
@@ -68,14 +68,66 @@ func printStackTree(repo *git.Repo, branches map[string]meta.Branch, currentBran
...
+	if branch.PullRequest != nil && branch.PullRequest.Permalink != "" {
+		branchInfo = branch.PullRequest.Permalink + " " + branchInfo
+	}
+
+	return branchInfo
+}
+
+// Check if branch is up to date with the parent branch.
+// This is doing `git diff <parentBranch> <givenBranch>`
+func getParentStatus(repo *git.Repo, branch meta.Branch) string {
```
draftcode, line 106: I thought that "need sync" would mean it needs to be rebased (== need av stack sync). In order to tell that, you would need to check if mergebase(parent, branch) == parent.

draftcode: Also, if you do that, it'd be nice to compare with the remote tracking branch for trunks. i.e. in a pseudo code:

```
if branch.Parent.Trunk {
    return mergebaseCheck("origin/" + branch.Parent.Name, branch.Name)
}
return mergebaseCheck(branch.Parent.Name, branch.Name)
```

doratzeng: ok updated to use mergebase. should i return empty string or "needs sync" in case of an error?

README.md
```diff
@@ -9,3 +9,4 @@ To run the command line:
 ```
 go run ./cmd/av [subcommand/flags...]
 ```
+testing testing
```
draftcode, line 12: Remove

cmd/av/stack_tree.go
draftcode: Usually better to return `(string, error)`

draftcode (changes requested):

cmd/av/stack_tree.go
```diff
@@ -68,14 +68,85 @@ func printStackTree(repo *git.Repo, branches map[string]meta.Branch, currentBran
...
+		return "", nil
+	}
+	
+	return "needs sync", nil
+}
+
+// Check if branch is up to date with the upstream branch.
+// This is doing `git diff <givenBranch> remotes/origin/<givenBranch>`
+func getUpstreamStatus(repo *git.Repo, branch meta.Branch) (string, error) {
+	upstreamExists, err := repo.DoesRemoteBranchExist(branch.Name)
```
doratzeng, line 136: note that this only checks the local reference for if the upstream branch exists. also if the branch has already been merged and deleted, we will not display the `not pushed` messaged.

discussed offline with @draftcode , I think this should be ok for now.

cmd/av/stack_tree.go
```diff
@@ -68,14 +68,85 @@ func printStackTree(repo *git.Repo, branches map[string]meta.Branch, currentBran
...
 		printStackTree(repo, branches, currentBranch, next, depth+1)
 	}
 }
+
+func getBranchInfo(repo *git.Repo, branch meta.Branch) string {
+	var branchInfo string
+
+	parentStatus, err := getParentStatus(repo, branch)
+	if err != nil {
+		fmt.Printf("<ERROR: could not get status of parent branch: %v>\n", err)
```
draftcode, line 91: This should return an error.

cmd/av/stack_tree.go
```diff
@@ -71,14 +71,89 @@ func printStackTree(repo *git.Repo, branches map[string]meta.Branch, currentBran
...
 		printStackTree(repo, branches, currentBranch, next, depth+1)
 	}
 }
+
+func getBranchInfo(repo *git.Repo, branch meta.Branch) (string, error) {
+	var branchInfo string
+
+	parentStatus, err := getParentStatus(repo, branch)
+	if err != nil {
+		fmt.Printf("<ERROR: could not get status of parent branch: %v>\n", err)
```
draftcode, line 98: No. Do this in the caller side.

In general, if a method looks like side-effect free (like this method), it should be side effect free.

draftcode (changes requested):

draftcode (approved):

Issue #78, opened 2023-04-12
Feature request: `av stack sync` should delete branches that are already merged


abatilo:
After I have `/aviator stack merge`'d my stack, the branches get merged and now if I run `av stack sync`, I get output that implies to me that I might need to re-open my branches:

```
⇒  av stack sync
Synchronizing branch abatilo/03-16-_TIDE-1490_Create_a_second_staging_EKS_cluster...
  - fetching latest pull request information for abatilo/03-16-_TIDE-1490_Create_a_second_staging_EKS_cluster
  - branch is a stack root, nothing to do (run av stack sync --trunk to sync against the latest commit in main)
  - WARNING: pull request #12903 is closed, skipping push
      - re-open the pull request (or create a new one with av pr create) to push changes


Synchronizing branch Install-post-create-CRDs...
  - fetching latest pull request information for Install-post-create-CRDs
  - already up-to-date with parent abatilo/03-16-_TIDE-1490_Create_a_second_staging_EKS_cluster
  - WARNING: pull request #13270 is closed, skipping push
      - re-open the pull request (or create a new one with av pr create) to push changes


Synchronizing branch Explicitly-list-terraform_actor-role...
  - fetching latest pull request information for Explicitly-list-terraform_actor-role
  - skipping sync for merged branch (merged in commit e798ae6)
```

This is a confusing UX if someone isn't paying enough attention, they might just follow these steps again.

PR #97, opened 2023-05-02
feat: Delete the merged branches in stack sync

draftcode:
This is a relatively large change. Previously, av stack sync didn't touch the
already merged branches. This changes this behavior and it always rebases the
merged branches. After the rebase, if it's sure that the changes are fully
merged (no leftovers that haven't merged yet), it automatically deletes the
branch and the child branches are reparented.

This addresses https://github.com/aviator-co/av/issues/78.

### DIFF
--- a/internal/actions/sync_branch.go
+++ b/internal/actions/sync_branch.go
@@ -15,7 +15,6 @@ import (
 	"github.com/aviator-co/av/internal/utils/ghutils"
 	"github.com/aviator-co/av/internal/utils/sliceutils"
 	"github.com/shurcooL/githubv4"
-	"github.com/sirupsen/logrus"
 )
 
 type SyncBranchOpts struct {
@@ -39,18 +38,15 @@ type SyncBranchResult struct {
 
 	// The updated branch metadata (if the rebase was successful)
 	Branch meta.Branch
+
+	// The branch should be deleted.
+	DeleteBranch bool
 }
 
 type (
 	SyncBranchContinuation struct {
 		// The original HEAD commit of the branch.
 		OldHead string `json:"oldHead"`
-		// The commit that we were rebasing the branch on top of.
-		ParentCommit string `json:"parentCommit"`
-
-		// If set, we need to re-assign the branch to be a stack root that is
-		// based on this trunk branch.
-		NewTrunk string `json:"newTrunk,omitempty"`
 	}
 )
 
@@ -95,17 +91,6 @@ func SyncBranch(
 			}
 		}
 
-		if branch.MergeCommit != "" {
-			_, _ = fmt.Fprint(os.Stderr,
-				"  - skipping sync for merged branch "+
-					"(merged in commit ", colors.UserInput(git.ShortSha(branch.MergeCommit)), ")"+
-					"\n",
-			)
-			return &SyncBranchResult{
-				RebaseResult: git.RebaseResult{Status: git.RebaseAlreadyUpToDate},
-			}, nil
-		}
-
 		var err error
 		res, err = syncBranchRebase(ctx, repo, tx, opts, branch)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


internal/actions/sync_branch.go
```diff
@@ -118,7 +103,25 @@ func SyncBranch(
...
 	}
 
-	if opts.Push {
+	if branch.PullRequest != nil && branch.PullRequest.State != githubv4.PullRequestStateOpen {
+		if lo, err := hasLeftover(repo, branch, opts); err != nil {
+			return nil, err
+		} else if lo {
+			if !opts.ToTrunk {
+				_, _ = fmt.Fprint(os.Stderr,
+					"  - ", colors.Warning("WARNING: The PR is closed, but without --trunk, we cannot tell if we can safely delete the branch."),
```
travigd, line 112: I don't understand this. Can you describe when this would happen?

internal/actions/sync_branch.go
```diff
@@ -347,21 +268,59 @@ func syncBranchRebase(
...
+	if branch.IsStackRoot() {
+		if opts.ToTrunk {
+			trunk := branch.Parent.Name
+			trunkHead, err := repo.RevParse(&git.RevParse{Rev: "origin/" + trunk})
+			if err != nil {
+				return false, errors.WrapIff(err, "failed to get HEAD of %q", trunk)
+			}
+			rebaseUpstream = trunkHead
+		} else if branch.MergeCommit == "" {
+			// The PR is closed without merge. There are two possibilities (1) the PR is
```
travigd, line 300: This comment is incorrect. We set `branch.MergeCommit` correctly when merged with FF mode.

See https://github.com/aviator-co/av/blob/f2240b0486022b968fcc0823edb1b99e2825cc2b/internal/actions/pr.go#L508 and https://github.com/aviator-co/av/blob/f2240b0486022b968fcc0823edb1b99e2825cc2b/internal/gh/pullrequest.go#L56-L65.

travigd: This also hinges on functionality in MergeQueue where we create the FF merge commit with the text "Closes #<number>" which generates the requisite timeline entry.

internal/actions/sync_branch.go
```diff
@@ -395,21 +354,12 @@ func syncBranchContinue(
...
 	case git.RebaseConflict:
 		msgRebaseResult(rebase)
-		return &SyncBranchResult{*rebase, opts.Continuation, branch}, nil
+		return &SyncBranchResult{*rebase, opts.Continuation, branch, false}, nil
 	default:
 		msgRebaseResult(rebase)
 	}
 
-	// Finish setting the new trunk for the branch
-	if opts.Continuation.NewTrunk != "" {
```
travigd, line 404: Is this not necessary anymore?

travigd (commented):
This is confusing. I don't fully understand. It seems rather complicated.

I don't think rebasing merged branches is necessary (or a particularly good idea -- we should avoid changing the users history in unexpected ways).

We should be able to determine whether or not there are commits that weren't merged by doing
```
# check if foo has un-pushed commits
git fetch origin foo
git rev-list origin/foo..foo
```

If the `rev-list` returns a non-empty list, then those are the commits that were present in `foo` that weren't in `origin/foo`. It's possibly that we'd want to replace the `origin/foo` with the HEAD commit of the PR (reported by the GH API) but it's still the same idea.

draftcode:
> This is confusing. I don't fully understand. It seems rather complicated.
> 
> I don't think rebasing merged branches is necessary (or a particularly good idea -- we should avoid changing the users history in unexpected ways).
> 
> We should be able to determine whether or not there are commits that weren't merged by doing
> 
> ```
> # check if foo has un-pushed commits
> git fetch origin foo
> git rev-list origin/foo..foo
> ```
> 
> If the `rev-list` returns a non-empty list, then those are the commits that were present in `foo` that weren't in `origin/foo`. It's possibly that we'd want to replace the `origin/foo` with the HEAD commit of the PR (reported by the GH API) but it's still the same idea.

I do not understand why un-pushed commits are relevant to merged commits...?

PR #98, opened 2023-05-02
feat: Add --skip option to av stack sync

draftcode:

### DIFF
--- a/cmd/av/stack_sync.go
+++ b/cmd/av/stack_sync.go
@@ -67,6 +67,8 @@ var stackSyncFlags struct {
 	Continue bool
 	// If set, abort an in-progress sync operation.
 	Abort bool
+	// If set, skip a commit and continue a previous sync.
+	Skip bool
 }
 
 var stackSyncCmd = &cobra.Command{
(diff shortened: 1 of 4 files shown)


cmd/av/stack_sync.go
```diff
@@ -141,23 +143,26 @@ base branch.
...
+		if stackSyncFlags.Continue || stackSyncFlags.Skip {
 			if state.CurrentBranch == "" {
 				return errors.New("no sync in progress")
 			}
 		} else {
-			// Not a --continue, we're trying to start a new sync from scratch.
+			// Not a --continue/--skip, we're trying to start a new sync from scratch.
 			if state.CurrentBranch != "" {
-				return errors.New("a sync is already in progress: use --continue or --abort")
+				return errors.New("a sync is already in progress: use --continue, --abort, or --skip")
```
travigd, line 165: I'd actually rather not suggest `--skip` to the average user. Generally people aren't going to want to do this very often and I think presenting it as an option at the same level as `--continue` or `--abort` will lead people to do it when they shouldn't.

travigd (approved):

cmd/av/stack_sync.go
draftcode: Sure. This is for addressing cases where merge conflict happens for the already merged branches. I usually use --skip because they are already merged, but I can remove --here. (Not sure what people should do for those cases otherwise.)

PR #99, opened 2023-05-02
refactor: Address golangci issues and introduce goimports

draftcode:
* Address io/ioutil deprecation. Replace it with os and io equivalents.
* Add goimports as import statements flips occasionally
* Replace deprecated structcheck with unused
* Run gofmt for all


draftcode:
> Should we add pre-commit hooks?

Maybe later.

PR #100, opened 2023-05-02
feat: Use remote tracking branch if it's trunk

draftcode:

### DIFF
--- a/internal/actions/reparent.go
+++ b/internal/actions/reparent.go
@@ -58,19 +58,26 @@ func Reparent(
 	}
 
 	// Check that the parent branch actually exists
-	parentSha, err := repo.RevParse(&git.RevParse{Rev: opts.NewParent})
+	parentBranch := opts.NewParent
+	if opts.NewParentTrunk {
+		parentBranch = "remotes/origin/" + opts.NewParent
+	}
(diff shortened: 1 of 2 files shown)


internal/actions/reparent.go
```diff
@@ -54,19 +54,26 @@ func Reparent(repo *git.Repo, opts ReparentOpts) (*ReparentResult, error) {
 	}
 
 	// Check that the parent branch actually exists
-	parentSha, err := repo.RevParse(&git.RevParse{Rev: opts.NewParent})
+	parentBranch := opts.NewParent
+	if opts.NewParentTrunk {
+		parentBranch = "remotes/origin/" + opts.NewParent
```
travigd, line 59: I'm not sure that this is what we want by default. It will be surprising to people that don't use the "no local `main` branch" workflow. Requiring users to pass `origin/main` explicitly will probably be better.

travigd: Actually I think I changed my mind on this. Assuming you always want to update on the latest version of trunk is probably reasonable. We can always add a `--no-fetch` type flag if anyone ever needs that.

travigd (approved):

PR #103, opened 2023-05-05
fix: Fix --trunk so that it actually rebases to the upstream trunk

draftcode:
Without this fix, the test fails.

### DIFF
--- a/internal/actions/sync_branch.go
+++ b/internal/actions/sync_branch.go
@@ -139,13 +139,48 @@ func syncBranchRebase(
 		return nil, errors.WrapIff(err, "failed to get head of branch %q", branch.Name)
 	}
 
-	if branch.IsStackRoot() {
-		trunk := branch.Parent.Name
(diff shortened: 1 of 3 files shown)


internal/actions/sync_branch.go
```diff
@@ -139,52 +139,71 @@ func syncBranchRebase(
 		return nil, errors.WrapIff(err, "failed to get head of branch %q", branch.Name)
 	}
 
-	if branch.IsStackRoot() {
-		trunk := branch.Parent.Name
-		if !opts.ToTrunk {
+	parent := branch.Parent
```
travigd, line 142: [minor] Can you make this `parentState` for disambiguation from `parentBranch`? I think it'll make it slightly easier to read below.

travigd (approved):

internal/actions/sync_branch.go
draftcode: Done.

PR #104, opened 2023-05-08
fix: Update recorded parent's HEAD after rebase

draftcode:

### DIFF
--- a/e2e_tests/helpers.go
+++ b/e2e_tests/helpers.go
@@ -4,6 +4,8 @@ import (
 	"testing"
 
 	"github.com/aviator-co/av/internal/git"
+	"github.com/aviator-co/av/internal/meta"
+	"github.com/aviator-co/av/internal/meta/jsonfiledb"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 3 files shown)


e2e_tests/helpers.go
```diff
@@ -12,3 +14,11 @@ func RequireCurrentBranchName(t *testing.T, repo *git.Repo, name string) {
 	require.NoError(t, err, "failed to determine current branch name")
 	require.Equal(t, name, currentBranch, "expected current branch to be %q, got %q", name, currentBranch)
 }
+
+func GetStoredParentBranchState(t *testing.T, repo *git.Repo, name string) meta.BranchState {
```
travigd, line 18: Not for this PR, but we should add a command to get this information that way we don't have to do this.

e2e_tests/helpers.go
```diff
@@ -12,3 +14,11 @@ func RequireCurrentBranchName(t *testing.T, repo *git.Repo, name string) {
...
+	// We shouldn't do this as part of an E2E test, but it's hard to ensure otherwise.
+	db, err := jsonfiledb.OpenRepo(repo)
```
travigd, line 20: Do we need to close this to close the file handle? (not sure if that's actually a part of the DB interface in hindsight... 🤔)

travigd (approved):
Thanks for finding/fixing this!!!!!!!

e2e_tests/helpers.go
draftcode: jsonfiledb doesn't have a file handle in it.

e2e_tests/helpers.go
draftcode: Sure

e2e_tests/helpers.go
travigd: Now that you say that I remember it lol. I wrote that code all of two weeks ago and remember none of it apparently.

PR #105, opened 2023-05-08
fix: Set the parent when creating a PR

draftcode:

### DIFF
--- a/internal/actions/pr.go
+++ b/internal/actions/pr.go
@@ -176,6 +176,7 @@ func CreatePullRequest(
 			Name:  defaultBranch,
 			Trunk: true,
 		}
+		branchMeta.Parent = parentState
 	}
 	prCompareRef := parentState.Name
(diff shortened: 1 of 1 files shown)


internal/actions/pr.go
```diff
@@ -176,6 +176,7 @@ func CreatePullRequest(
 			Name:  defaultBranch,
 			Trunk: true,
 		}
+		branchMeta.Parent = parentState
```
jainankit, line 179: do you know why the `parentState` would be empty here to begin with? Just want to make sure we understand when does this particularly happen.

draftcode: The parent state is empty when the branch is not created with `av`, like `git checkout -b topic`.

jainankit: Can this also happen when an associated branch was deleted after merging a parent PR? I think that's what happened in my case.

jainankit (approved):

PR #106, opened 2023-05-08
fix: Cleanup on failure for av stack branch-commit

### LINKED ISSUE #102: If a av stack branch-commit command fails, you cannot rerun it
If `av stack branch-commit` command fails because of a pre-commit failure, since the branch is already created, rerunning the command fails.

## Expected
If the command fails, it would be ideal to rollback to the old branch and delete the new branch. That way the command can be replayed after fixing the pre-commit error.

travigd:
Deletes a branch during `av stack branch-commit` if the commit fails (e.g., due to pre-commit hooks).

Also a few minor things like language tweaks and adding `av stack bc` as an alias.

Fixes #102.

### DIFF
--- a/cmd/av/stack_branchcommit.go
+++ b/cmd/av/stack_branchcommit.go
@@ -1,7 +1,10 @@
 package main
 
 import (
+	"fmt"
 	"github.com/aviator-co/av/internal/utils/cleanup"
+	"github.com/aviator-co/av/internal/utils/colors"
+	"os"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


cmd/av/stack_branchcommit.go
```diff
@@ -43,8 +46,9 @@ var stackBranchCommitFlags struct {
 
 var stackBranchCommitCmd = &cobra.Command{
 	Use:          "branch-commit [flags]",
-	Short:        "create a new stacked branch and a commit",
-	Long:         "Create a new branch that is stacked on the current branch, and call git-commit with the specified arguments.",
+	Aliases:      []string{"bc"},
+	Short:        "switch to a new stacked branch commit staged changes to it",
```
jainankit, line 50: This is very confusing right now.
```suggestion
	Short:        "create a new stacked branch and commit staged changes to it",
```

jainankit (approved):

PR #108, opened 2023-05-10
fix: Address the case where the middle branch is merged to parent

draftcode:


travigd (approved):
Woo! Can you also add a message that says something to the effect of
```
  - Parent branch <parent> was merged into non-trunk branch <grandparent>, reparenting <child> onto <grandparent>
```

draftcode:
> 

Done

PR #111, opened 2023-05-12
fix: Update the parent consistently

draftcode:
When a conflict happens, in some cases the parent is updated before
conflict is resolved. When sync is aborted, the parent is already moved,
so it won't get back to the original state. The parent update should be
done after --continue.

Also, the parent should be updated even if it's not trunk. It should
update the parent hash as well.

This change creates a sync continuation for all cases. The continuation
contains the data needed for updating parents. The continuation is
called inline when there's no conflict. If there's a conflict, the
continuation is saved and applied later.

### DIFF
--- a/internal/actions/sync_branch.go
+++ b/internal/actions/sync_branch.go
@@ -15,7 +15,6 @@ import (
 	"github.com/aviator-co/av/internal/utils/ghutils"
 	"github.com/aviator-co/av/internal/utils/sliceutils"
 	"github.com/shurcooL/githubv4"
-	"github.com/sirupsen/logrus"
 )
 
 type SyncBranchOpts struct {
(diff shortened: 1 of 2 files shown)


internal/actions/sync_branch.go
```diff
@@ -435,29 +403,29 @@ func syncBranchContinue(
...
 	return &SyncBranchResult{*rebase, nil, branch}, nil
 }
 
-func syncBranchUpdateNewTrunk(
+func syncBranchUpdateParent(
 	tx meta.WriteTx,
 	branch meta.Branch,
-	newTrunk string,
+	continuation *SyncBranchContinuation,
 ) (meta.Branch, error) {
```
draftcode, line 418: Actually, this has a bug. I'll fix that. We should addremove the child/parents properly.

draftcode: Fixed. Now this adds a child when the parent is changed and the parent is not trunk.

internal/actions/sync_branch.go
```diff
@@ -43,14 +42,10 @@ type SyncBranchResult struct {
 
 type (
 	SyncBranchContinuation struct {
```
travigd, line 44: Maybe makes sense to just rename this to `SyncBranchState` at this point?

travigd (approved):

internal/actions/sync_branch.go
draftcode: I think continuation makes sense. It's indeed the continuation post rebase.

PR #113, opened 2023-05-15
feat: Save PR message if creation fails

travigd:
Saves the pull request description (entered via editor) to a temporary file if creating the pull request fails. It will re-use the title/description if the user tries to re-create the PR (also just logs the file path to the console in case they want that).

The most common cause of this error (in my experience) is expired auth tokens and/or the base branch of a stack having been deleted on GitHub. (we should provide a better experience for the latter but that's outside the scope of this PR)

### DIFF
--- a/internal/actions/pr.go
+++ b/internal/actions/pr.go
@@ -5,6 +5,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/aviator-co/av/internal/utils/cleanup"
+	"github.com/aviator-co/av/internal/utils/sanitize"
 	"os"
 	"path/filepath"
 	"strings"
@@ -129,6 +131,9 @@ func CreatePullRequest(
 		return nil, ErrRepoNotInitialized
 	}
 
+	var cu cleanup.Cleanup
+	defer cu.Cleanup()
+
 	_, _ = fmt.Fprint(os.Stderr,
 		"Creating pull request for branch ", colors.UserInput(opts.BranchName), ":",
 		"\n",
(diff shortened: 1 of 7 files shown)


internal/actions/pr.go
```diff
@@ -234,8 +235,25 @@ func CreatePullRequest(
 			commits = append(commits, *commit)
 		}
 
-		// We need to open an editor to ask the user. Try to populate with
-		// reasonable defaults here.
+		// If a saved pull request description exists, use that.
+		saveFile := filepath.Join(os.TempDir(), fmt.Sprintf("av-pr-%s.md", sanitize.FileName(opts.BranchName)))
```
draftcode, line 239: Can we use avDir https://github.com/aviator-co/av/blob/7d8e52eff86de083a948426f91d24b110147b479/cmd/av/stack_sync.go#L351 instead of the TempDir. tmpfs is ephemeral.

travigd: I think tempdir makes sense here. It's decently ephemeral data and it's also possible that if they fail to create a PR using av they'll just create it on GH and we'll never clean up that old data. It's not liable to be more than a few kb for most people but still I think tempfs makes sense here.

draftcode: Ephemeral here means it won't survive after reboot. When a machine has a network issue and has to reboot, after reboot the message is lost.

travigd: Yeah, I think that's a fine trade off to make here. Most people don't reboot very often and it's likely (at least I imagine) that if you fail to open a PR you'll deal with it before the next time you'd reboot your machine.

draftcode (approved):

PR #117, opened 2023-05-16
fix: Handle the original parent is a trunk case

draftcode:

### DIFF
--- a/internal/actions/sync_branch.go
+++ b/internal/actions/sync_branch.go
@@ -179,12 +179,21 @@ func syncBranchRebase(
 			}, nil
 		}
 
+		var origUpstream string
+		if origParent.Trunk {
+			// We do not know the original trunk commit hashes. Use the current one as
+			// an approximation.
+			origUpstream = newUpstreamCommitHash
+		} else {
+			origUpstream = origParent.Head
+		}
+
 		continuation := SyncBranchContinuation{
 			NewParentName: parentState.Name,
 		}
 		rebase, err := repo.RebaseParse(git.RebaseOpts{
 			Branch:   branch.Name,
-			Upstream: origParent.Head,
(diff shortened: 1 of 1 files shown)


internal/actions/sync_branch.go
```diff
@@ -248,6 +257,16 @@ func syncBranchRebase(
 		// Note that we've introduced B into the history of stacked-2, but
 		// not C or D since those commits come after M.
 		newUpstreamCommitHash := origParentBranch.MergeCommit
+		var origUpstream string
+		if origParent.Trunk {
+			var err error
+			origUpstream, err = repo.RevParse(&git.RevParse{Rev: "origin/" + origParent.Name})
```
jainankit, line 263: for my understanding, if i read the `BranchState` correctly, the `BranchState.Head` represents the SHA of the parent branch, not the SHA of the current branch. But `git rev-parse` gives you SHA of the current branch.

Is that an intended change?

internal/actions/sync_branch.go
```diff
@@ -179,12 +179,21 @@ func syncBranchRebase(
 			}, nil
 		}
 
+		var origUpstream string
+		if origParent.Trunk {
+			// We do not know the original trunk commit hashes. Use the current one as
+			// an approximation.
+			origUpstream = newUpstreamCommitHash
```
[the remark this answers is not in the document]
draftcode: `newUpstreamCommitHash` is either a result of the revparse or the merge commit which is close to the original upstream.

internal/actions/sync_branch.go
draftcode: BranchState.Head is empty if Trunk = true. Hence this change.

jainankit (approved):

PR #121, opened 2023-05-18
[MER-2352] Make branch adoption work

### LINKED ISSUE #120: Reparenting won't work
```
% av stack sync --parent main
  - Re-parenting branch mytopic onto main
      - ERROR: conflict while rebasing branch mytopic onto main
Failed to re-parent branch: resolve the conflicts and continue the sync with av stack sync --continue
hint:
    fatal: invalid upstream ''
```

draftcode:
Close #120.

### DIFF
--- a/internal/actions/reparent.go
+++ b/internal/actions/reparent.go
@@ -34,6 +34,24 @@ func Reparent(
 	tx meta.WriteTx,
 	opts ReparentOpts,
 ) (*ReparentResult, error) {
+	branchMeta, exist := tx.Branch(opts.Branch)
+	if !exist {
+		_, _ = fmt.Fprint(os.Stderr,
+			"  - Adopting a branch ", colors.UserInput(opts.Branch), " to Av", colors.UserInput(opts.NewParent),
+			"\n",
+		)
+		branchMeta.Parent.Name = opts.NewParent
+		branchMeta.Parent.Trunk = opts.NewParentTrunk
+		if !branchMeta.Parent.Trunk {
+			head, err := repo.RevParse(&git.RevParse{Rev: opts.NewParent})
+			if err != nil {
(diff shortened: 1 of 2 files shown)


internal/actions/reparent.go
```diff
@@ -34,6 +34,24 @@ func Reparent(
 	tx meta.WriteTx,
 	opts ReparentOpts,
 ) (*ReparentResult, error) {
+	branchMeta, exist := tx.Branch(opts.Branch)
+	if !exist {
```
jainankit, line 38: so this was not working previously if the current branch was not part of any existing stack?

jainankit (approved):

travigd (commented):
I'm a bit confused by this -- can you hold off on merging for a second? I wanna look at it a bit more if that's okay!

internal/actions/reparent.go
```diff
@@ -34,6 +34,24 @@ func Reparent(
...
+		branchMeta.Parent.Name = opts.NewParent
+		branchMeta.Parent.Trunk = opts.NewParentTrunk
+		if !branchMeta.Parent.Trunk {
+			head, err := repo.RevParse(&git.RevParse{Rev: opts.NewParent})
+			if err != nil {
+				return nil, errors.WrapIff(err, "failed to get HEAD of %q", opts.NewParent)
+			}
+			branchMeta.Parent.Head = head
+		}
+		tx.SetBranch(branchMeta)
```
travigd, line 52: I don't think we need to do this `SetBranch` here.

I think it's effectively a no-op right now because we do the `SetBranch` in `reparentWriteMetadata` which is called on every code path (unless there's a conflict... is that an issue? 🙃).

At any rate, I think this is fine (as in it ~~works~~) but it's a little hard to follow.

travigd (approved):

PR #122, opened 2023-05-19
feat: Adding av commit create

ohcnivek:
### feature: av commit create
Screen Shot 2023-05-19 at 3 11 01 PM

- need to/ will add more flags on top of this pr, but this is the scaffolding for the commit create cmd

### DIFF
--- a/cmd/av/commit_create.go
+++ b/cmd/av/commit_create.go
@@ -70,7 +70,7 @@ func init() {
 		return errExitSilently{1}
 	}
 
-	state, err := readStackSyncState(repo)
+	state, err := actions.ReadStackSyncState(repo)
 	if err != nil && !os.IsNotExist(err) {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


cmd/av/commit_create.go
```diff
@@ -0,0 +1,126 @@
...
+		ExitError:   true,
+		Interactive: true,
+	}); err != nil {
+		_, _ = fmt.Fprint(os.Stderr,
+			"\n", colors.Failure("Failed to create commit."), "\n",
+		)
+		return errExitSilently{1}
+	}
+
+	state, err := readStackSyncState(repo)
```
travigd, line 73: A lot of this logic looks like its copied from `av stack sync`. We probably shouldn't have two different implementations of this logic.

It might be necessary to move the stack sync logic into a function so we can re-use that logic (maybe `actions.SyncStack`?).

cmd/av/commit_create.go
```diff
@@ -0,0 +1,101 @@
...
+var commitCreateCmd = &cobra.Command{
+	Use:   "create",
+	Short: "create a commit",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		repo, err := getRepo()
+		if err != nil {
+			return err
+		}
+
+		currentBranchName, _ := repo.CurrentBranchName()
```
draftcode, line 35: Capture an error.

draftcode (approved):

PR #123, opened 2023-05-19
feat(stack reorder): Add command scaffolding

travigd:
Breaking this into a few parts.

This PR is intended to just define the base scaffolding. Namely, we're just defining the structure of the `stack-branch` and `pick` commands (ultimately the goal is to support all the commands from `git rebase --interactive` in addition to some av specific ones like `stack-branch`).

This code is not called from anywhere currently.

### DIFF
--- a/internal/reorder/stackbranch.go
+++ b/internal/reorder/stackbranch.go
@@ -7,7 +7,7 @@ import (
 
 // StackBranchCmd is a command to create a new branch in a stack.
 //
-//	branch <branch-name> [--parent <parent-branch-name>] [--trunk <trunk-branch-name>]
+//	stack-branch <branch-name> [--parent <parent-branch-name>] [--trunk <trunk-branch-name>]
 type StackBranchCmd struct {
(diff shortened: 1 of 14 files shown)


internal/reorder/stackbranch.go
```diff
@@ -0,0 +1,61 @@
+package reorder
+
+import (
+	"github.com/spf13/pflag"
+	"strings"
+)
+
+// StackBranchCmd is a command to create a new branch in a stack.
+//
+//	branch <branch-name> [--parent <parent-branch-name>] [--trunk <trunk-branch-name>]
```
draftcode, line 10: stack-branch

draftcode (approved):

internal/reorder/stackbranch.go
travigd: Yep, thanks. Originally it was named just `branch` (before the decision to make it `stack-branch` to avoid conflicts with rebase's `break`) and I just missed these. Thanks for catching!

PR #126, opened 2023-05-24
feat(stack reorder): Implement PickCmd

travigd:
Implements the actual `pick` command (along with adding the necessary other things like `Repo.CherryPick`).

### DIFF
--- a/internal/git/cherrypick.go
+++ b/internal/git/cherrypick.go
@@ -1,0 +1,96 @@
+package git
+
+import (
+	"emperror.dev/errors"
+	"fmt"
+	"strings"
+)
+
+type CherryPickResume string
+
+const (
+	CherryPickContinue CherryPickResume = "continue"
+	CherryPickSkip     CherryPickResume = "skip"
+	CherryPickQuit     CherryPickResume = "quit"
+	CherryPickAbort    CherryPickResume = "abort"
+)
+
+type CherryPick struct {
+	// Commits is a list of commits to apply.
+	Commits []string
+
+	// NoCommit specifies whether or not to cherry-pick without committing
+	// (equivalent to the --no-commit flag on `git cherry-pick`).
+	NoCommit bool
+
+	// FastForward specifies whether or not to fast-forward the current branch
+	// if possible (equivalent to the --ff flag on `git cherry-pick`).
+	// If true, and the parent of the commit is the current HEAD, the HEAD
+	// will be fast forwarded to the commit (instead of re-applied).
+	FastForward bool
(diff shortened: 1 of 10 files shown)


internal/reorder/pick.go
```diff
@@ -10,12 +15,37 @@ type PickCmd struct {
...
 
-func (b PickCmd) Execute(ctx *Context) error {
-	panic("not implemented")
+func (p PickCmd) Execute(ctx *Context) error {
+	err := ctx.Repo.CherryPick(git.CherryPick{
+		Commits: []string{p.Commit},
+		// Use FastForward to avoid always amending commits.
+		FastForward: true,
+	})
+	if conflict, ok := errutils.As[git.ErrCherryPickConflict](err); ok {
```
travigd, line 24: Not sure if this is super bad practice. The alternative is doing
```go
var conflict git.ErrCherryPickConflict
if errors.As(err, &conflict) {
```
which is only one extra line.

Let me know if this burns your eyes to look at and I'll change it 😅.

internal/git/cherrypick.go
```diff
@@ -0,0 +1,96 @@
...
+	FastForward bool
+
+	// Resume specifies how to resume a cherry-pick operation that was
+	// interrupted by a conflict (equivalent to the --continue, --skip, --quit,
+	// and --abort flags on `git cherry-pick`).
+	// Mutually exclusive with all other options.
+	Resume CherryPickResume
+}
+
+type CherryPickResult struct {
```
draftcode, line 39: I guess initially thinking about returning this, but decided not to use this?

internal/reorder/pick.go
draftcode: Really? I feel like I do this assign-in-if frequently to limit the scope. (Also mentioned in https://go.dev/doc/effective_go#if)

draftcode (approved):

internal/reorder/pick.go
travigd: Yeah, the only difference here is that we have to declare the concrete type of the error which isn't possible without using generics here. There's no
```go
if var conflict git.ErrCherryPickConflict; errors.As(err, &conflict) { ... }
```
construct in Go.

I guess I could also do
```go
if conflict := git.ErrCherryPickConflict{}; errors.As(err, &conflict) { ... }
```
but I've never seen that usage before.

internal/git/cherrypick.go
travigd: Yep, good catch.

PR #134, opened 2023-05-25
[MER-2351] Add --all option to sync all branches

draftcode:

### DIFF
--- a/cmd/av/stack_sync.go
+++ b/cmd/av/stack_sync.go
@@ -215,6 +215,19 @@ base branch.
 			// we try to do something in the repo before we finish that)
 			branchesToSync = []string{state.CurrentBranch}
 			state.Branches = branchesToSync
+		} else if syncFlags.All {
+			for _, br := range tx.AllBranches() {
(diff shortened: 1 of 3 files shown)


cmd/av/stack_sync.go
```diff
@@ -251,6 +264,10 @@ base branch.
 }
 
 func init() {
+	stackSyncCmd.Flags().BoolVar(
+		&syncFlags.All, "all", false,
+		"synchronize all branches",
+	)
```
travigd, line 270: Can you add a mutually exclusive flag with `--current`? I recently learned this is a thing within Cobra: https://github.com/aviator-co/av/pull/133/files#diff-148b0d72c2cfeab9cfe0849ff12194c92c9636b23de06402645f2455a9a6d653R222.

travigd (approved):

cmd/av/stack_sync.go
draftcode: TIL. Done.

PR #136, opened 2023-05-26
refactor(stack sync): Don't pass StackSyncFlags to actions.SyncStack

travigd:
Just found this bug while doing some stuff locally.

### DIFF
--- a/internal/actions/pr.go
+++ b/internal/actions/pr.go
@@ -633,7 +633,7 @@ func ParsePRMetadata(
 		return
 	}
 
-	// This will skipNextCommit over any data lines (since those weren't consumed by buf,
+	// This will skip over any data lines (since those weren't consumed by buf,
 	// only by the copy of buf).
 	if err := readLineUntil(buf, "```\n"); err != nil {
 		reterr = errors.WrapIff(err, "expecting closing \"```\"")
--- a/internal/actions/pr.go
+++ b/internal/actions/pr.go
@@ -633,7 +633,7 @@ func ParsePRMetadata(
 		return
 	}
 
-	// This will skip over any data lines (since those weren't consumed by buf,
+	// This will skipNextCommit over any data lines (since those weren't consumed by buf,
 	// only by the copy of buf).
 	if err := readLineUntil(buf, "```\n"); err != nil {
 		reterr = errors.WrapIff(err, "expecting closing \"```\"")
--- a/internal/actions/push.go
+++ b/internal/actions/push.go
@@ -27,7 +27,7 @@ type PushOpts struct {
 	Force ForceOpt
 	// If true, require the corresponding branch exist on the remote (otherwise, don't push).
 	SkipIfRemoteBranchNotExist bool
-	// If true, skip pushing the branch if the corresponding branch on the remote points to the
+	// If true, skipNextCommit pushing the branch if the corresponding branch on the remote points to the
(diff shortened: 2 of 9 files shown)


ohcnivek (approved):

draftcode:
If we do this, shouldn't we remove the individual fetches?

travigd:
Oh hmmm... I think this actually is not the right fix. I think as part of @ohcnivek's refactor in #122, we no longer read the flags correctly in `StackSync` (we take it from the state instead of the flags). I can fix that though.

internal/actions/sync_stack.go
```diff
@@ -69,17 +56,35 @@ type StackSyncState struct {
...
 
-// Performs stack sync on all branches in branchesToSync.
+type (
+	SyncStackOpt  func(*syncStackOpts)
+	syncStackOpts struct {
+		skipNextCommit bool
+	}
+)
+
+func WithSkipNextCommit() SyncStackOpt {
```
travigd, line 66: I'm not 100% on this but I thought doing the functional option style was a better choice than putting it in the state (since the state gets persisted to disk, but `--skip` is fundamentally a one-time thing.

internal/actions/reparent.go
```diff
@@ -149,7 +149,7 @@ func ReparentSkipContinue(
 	}
 
 	if output.ExitCode != 0 && strings.Contains(string(output.Stderr), "no rebase in progress") {
-		// If there's no rebase, assume the user did `git rebase --continue/--skip` manually.
+		// If there's no rebase, assume the user did `git rebase --continue/--skipNextCommit` manually.
```
draftcode, line 152: There's no such option like `--skipNextCommit`.

internal/actions/pr.go
```diff
@@ -633,7 +633,7 @@ func ParsePRMetadata(
 		return
 	}
 
-	// This will skip over any data lines (since those weren't consumed by buf,
+	// This will skipNextCommit over any data lines (since those weren't consumed by buf,
```
draftcode: I guess this is a mistake.

internal/actions/push.go
```diff
@@ -27,7 +27,7 @@ type PushOpts struct {
 	Force ForceOpt
 	// If true, require the corresponding branch exist on the remote (otherwise, don't push).
 	SkipIfRemoteBranchNotExist bool
-	// If true, skip pushing the branch if the corresponding branch on the remote points to the
+	// If true, skipNextCommit pushing the branch if the corresponding branch on the remote points to the
```
draftcode: This is a mistake.

internal/actions/sync_branch.go
```diff
@@ -24,7 +24,9 @@ type SyncBranchOpts struct {
 	// If specified, synchronize the branch against the latest version of the
 	// trunk branch. This value is ignored if the branch is not a stack root.
 	ToTrunk bool
-	Skip    bool
+	// If true, skipNextCommit the current commit.
```
draftcode: This is a mistake. s/skipNextCommit/skip/

internal/actions/sync_stack.go
draftcode: Using this style seems to me too much for `internal` package, but if you want to use this sure.

draftcode (changes requested):

draftcode (approved):

PR #142, opened 2023-05-31
refactor: Remove Children from BranchMeta

travigd:
Originally, we stored the children in the `meta.Branch` struct. This resulted in a duplication of the parent/child relationship (we store the parent of each branch in `Branch.Parent` as well as the children of each branch in `Branch.Children`).

This was because we had to read Git refs for each branch to get its metadata which is expensive. Now that we store the state of the repo in-memory, it's pretty trivial to just reconstruct the branch dependency graph whenever we need it.

### DIFF
--- a/internal/meta/branch.go
+++ b/internal/meta/branch.go
@@ -2,7 +2,7 @@ package meta
 
 import (
 	"encoding/json"
-	"fmt"
+	"golang.org/x/exp/slices"
 
 	"emperror.dev/errors"
 	"github.com/shurcooL/githubv4"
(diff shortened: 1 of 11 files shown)


internal/meta/branch.go
```diff
@@ -152,23 +143,27 @@ func Trunk(tx ReadTx, name string) (string, bool) {
 	return "", false
 }
 
-func RebuildChildren(tx WriteTx) {
+// Children returns all the immediate children of the given branch.
+func Children(tx ReadTx, name string) []Branch {
 	branches := tx.AllBranches()
```
travigd, line 148: This creates a copy of the map so is probably kind of expensive, but also I don't think it matters at all given that there's essentially zero chance that any reasonable user will ever surpass 1000 branches.

draftcode (approved):

PR #144, opened 2023-05-31
refactor: Delete unused code

draftcode:


internal/reorder/cmds.go
```diff
@@ -3,9 +3,10 @@ package reorder
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/aviator-co/av/internal/meta"
 	"io"
 
+	"github.com/aviator-co/av/internal/meta"
+
```
travigd, line 9: why the newlines? 🤔 

also I ran the `pre-commit` on all the files, wonder why this wasn't caught.

travigd (approved):

internal/reorder/cmds.go
draftcode: I'm actually not sure...

PR #145, opened 2023-06-01
feat: [MER-2339] add av auth status

doratzeng:


draftcode (commented):
Can you install pre-commit and try formatting files?

internal/gql/client.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"context"
+
+	"github.com/aviator-co/av/internal/config"
+	"github.com/shurcooL/graphql"
+	"golang.org/x/oauth2"
+)
+
+func GraphQLClient() *graphql.Client {
```
travigd, line 11: In Go, we always use the package name to qualify the names within that package, so the `GraphQL` in this function name is redundant with the `gql` of the package name.

```suggestion
func NewClient() *graphql.Client {
```
might be a more idiomatic name.

internal/config/config.go
```diff
@@ -88,4 +93,8 @@ func loadFromEnv() {
 	} else if githubToken := os.Getenv("GITHUB_TOKEN"); githubToken != "" {
 		Av.GitHub.Token = githubToken
 	}
+
+	if apiToken := os.Getenv("AVIATOR_APITOKEN"); apiToken != "" {
```
travigd, line 97: Can we do 
```suggestion
	if apiToken := os.Getenv("AV_API_TOKEN"); apiToken != "" {
```
instead?

cmd/av/auth_status.go
```diff
@@ -0,0 +1,44 @@
...
+			fmt.Printf("<ERROR: %v>\n", err)
+			return err
+		}
+
+		if query.Viewer.Email == "" {
+			fmt.Fprintln(
+				os.Stderr,
+				"Error: Could not find auth info, please verify that your API Token is correct.",
+			)
+			return nil
```
travigd, line 38: You can return `ErrExitSilently{1}` here that way the process exits with a non-zero exit code.

travigd (commented):
Left a bunch of comments. It's nothing major, just focusing being consistent with other things in the CLI and making sure we have a level of polish for this kind of thing since it's directly end-user facing.

internal/avgql/client.go
```diff
@@ -0,0 +1,19 @@
...
+	"context"
+	"fmt"
+
+	"github.com/aviator-co/av/internal/config"
+	"github.com/shurcooL/graphql"
+	"golang.org/x/oauth2"
+)
+
+func NewClient() *graphql.Client {
+	fmt.Println("token: ", config.Av.Aviator.APIToken)
```
draftcode, line 13: Debug statement?

draftcode (approved):

internal/avgql/client.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/shurcooL/graphql"
+	"golang.org/x/oauth2"
+)
+
+func NewClient() *graphql.Client {
+	src := oauth2.StaticTokenSource(
+		&oauth2.Token{AccessToken: config.Av.Aviator.APIToken},
+	)
+	httpClient := oauth2.NewClient(context.Background(), src)
+	return graphql.NewClient("https://api.aviator.co/graphql", httpClient)
```
travigd, line 16: Can you make this configurable? This will probably be useful for testing things locally.

```suggestion
    apiURL := os.Getenv("AV_GRAPHQL_URL")
    if apiURL == "" { apiURL = "https://api.aviator.co/graphql" }
	return graphql.NewClient(apiURL, httpClient)
```

travigd (approved):
💯

PR #150, opened 2023-06-01
refactor: Fix the import lines

draftcode:
Looks like there is a bug in goimports that creates multiple blocks.

There is a related bug report when used with -local, which we don't use,
but somehow observed a similar effect.
https://github.com/golang/go/issues/47440

Anyway, rerunning goimports won't split the blocks. This change keeps
the import blocks as described in
https://github.com/golang/go/wiki/CodeReviewComments#imports.

### DIFF
--- a/cmd/av/fetch.go
+++ b/cmd/av/fetch.go
@@ -5,13 +5,12 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/aviator-co/av/internal/actions"
-	"github.com/aviator-co/av/internal/utils/cleanup"
-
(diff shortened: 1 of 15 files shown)


travigd:
Is there a way we can prevent this from happening in the future? e.g., with a flag to `goimports`?

jainankit (approved):

draftcode:
> Is there a way we can prevent this from happening in the future? e.g., with a flag to `goimports`?

idk. Haven't figure out the condition.

PR #152, opened 2023-06-01
doc: [MER-2383] Add manpage for av stack sync

draftcode:

### DIFF
--- a/docs/av-stack-sync.1.md
+++ b/docs/av-stack-sync.1.md
@@ -1,0 +1,92 @@
+# av-stack-sync 1 "" av-cli "Aviator CLI User Manual"
+
+# NAME
+
+av-stack-sync - Synchronize stacked branches
+
+# SYNOPSIS
+
+`` av stack sync [--all | --current] [--no-push] [--no-fetch] [--prune]
+              [--trunk] [--continue | --abort | --skip] [--parent=<parent>]
+``
+
+# DESCRIPTION
+
+Over the time, branches get unsynchronized in many ways. Some branches are
+merged. The upstream branch is moved. A parent branch get their commit amended.
+`av stack sync` synchronizes the branches following the changes.
+
+For each branch, this command does the following:
+
+* Rebase onto the parent branch. By default, if the parent is the trunk branch
+  (e.g. `main`), this step is skipped. If `--trunk` is used, it fetches the
+  trunk branch from the remote and rebase onto it.
+
+* Push to the remote branch. With Git's default config, the push updates the
+  same name branch on the remote.
+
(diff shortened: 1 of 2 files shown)


docs/av-stack-sync.1.md
```diff
@@ -0,0 +1,61 @@
+# av-stack-sync 1 "" av-cli "Aviator CLI User Manual"
+
+# NAME
+
+av-stack-sync - Synchronize stacked branches
```
jainankit, line 5:
```suggestion
av-stack-sync - Synchronize stacked branches from their parent branches
```

docs/av-stack-sync.1.md
```diff
@@ -0,0 +1,61 @@
...
+              [--trunk] [--continue | --abort | --skip] [--parent=<parent>]
+``
+
+# DESCRIPTION
+
+Synchronize stacked branches to be up-to-date with their parent branches.
+
+By default, this command will sync all branches starting at the root of the
+stack and recursively rebasing each branch based on the latest commit from the
+parent branch.
```
jainankit, line 19: please also add a line to describing that this does not update references from remote, but pushes the new references to remote.

docs/av-stack-sync.1.md
```diff
@@ -0,0 +1,61 @@
...
+
+If the --current flag is given, this command will not recursively sync dependent
```
jainankit, line 21: can we start explaining this with what this does vs what it does not do:

```
If the --current flag is given, this command will only sync stack up to the current branch starting from the first branch in the stack. It will not recursively sync ...
```

docs/av-stack-sync.1.md
```diff
@@ -0,0 +1,61 @@
...
+branches of the current branch within the stack. This allows you to make changes
+to the current branch before syncing the rest of the stack.
+
+If the --trunk flag is given, this command will synchronize changes from the
```
jainankit, line 25: I wonder if it's worth also flagging this in the description on line 17.
```
By default, this command does not synchronize changes from the mainline branch of the repository.
```

docs/av-stack-sync.1.md
draftcode: Discussed offline, but we do not use "synchronize from".

docs/av-stack-sync.1.md
draftcode: Done.

docs/av-stack-sync.1.md
draftcode: Done.

docs/av-stack-sync.1.md
draftcode: Mentioned.

jainankit (approved):

PR #153, opened 2023-06-01
[MER-2347] No positional args.

ErichKramer:
Has no test code.

Arguably this warrants adding a test case that verifies we error on positional args to stack sync. 

I looked at our e2e_tests for a bit and didn't have an immediate idea of where/ how to add this and I'm not yet familiar with gittest. Feel free to bounce back to me that this needs a testcase.


travigd (approved):
Minor enough that it's not an issue to add tests imo. FYI gittest is just an internal utility library to create temporary git repos for testing on.

cmd/av/stack_sync.go
```diff
@@ -48,6 +48,7 @@ latest commit to the repository base branch (e.g., main or master) into the
 stack. This is useful for rebasing a whole stack on the latest changes from the
 base branch.
 `),
+	PositionalArgs: cobra.NoArgs,
```
draftcode, line 51: This should be `Args`

draftcode (approved):

cmd/av/stack_sync.go
ErichKramer: Yes, I had this locally and then pushed the wrong state. >.<
Rewriting history please hold.

PR #155, opened 2023-06-02
doc: Add manpage for av stack branch and fix av stack tidy

jainankit:

### DIFF
--- a/docs/av-stack-branch.1.md
+++ b/docs/av-stack-branch.1.md
@@ -1,0 +1,28 @@
+# av-stack-branch 1 "" av-cli "Aviator CLI User Manual"
+
+# NAME
+
+av-stack-branch - Create or rename a branch in the stack
+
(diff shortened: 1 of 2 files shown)


docs/av-stack-branch.1.md
```diff
@@ -0,0 +1,28 @@
...
+
+`av stack branch [-m | --rename] [--parent <parent_branch>] <branch-name>`
+
+# DESCRIPTION
+
+Create a new branch that is stacked on the current branch by default
+
+If the --rename/-m flag is given, the current branch is renamed to the name
+instead of creating a new branch. Branches should only be renamed
+with this command (not with git branch -m ...) because av needs to update
```
draftcode, line 17: Now realizing that we should have a way to retrofit the renames that happen outside of av. People would just do that, and later want to fix that.

draftcode (approved):

PR #157, opened 2023-06-05
feat(stack reorder): Generate plan for stack

travigd:
Getting pretty close to having an MVP on this feature.

The generated plan in the test case looks like
```
    plan_test.go:57: stack-branch one --trunk main@a50ac627fb10b6a7e19d39fde5f925ff9d2e9c29
    plan_test.go:57: pick 51f48a6  # Write file
    plan_test.go:57: pick baf3979  # Write file
    plan_test.go:57: stack-branch two --parent one
    plan_test.go:57: pick d856653  # Write fichier
    plan_test.go:57: pick 28fbcea  # Write fichier
```

The test cases for plan are a bit brittle at the moment, but I think we're still in the stage of figuring it out such that I think that's fine for now.


internal/git/catfile.go
```diff
@@ -69,9 +76,14 @@ func (r *Repo) GetRefs(opts *GetRefs) ([]*GetRefsItem, error) {
 			return nil, errors.Wrap(err, "failed to read cat-file output")
 		}
 		item.Contents = make([]byte, size)
-		_, err = io.ReadFull(output, item.Contents)
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to read cat-file output")
+		if n, err := io.ReadFull(output, item.Contents); err != nil {
+			logrus.WithFields(logrus.Fields{
```
travigd, line 80: Ended up needing to debug this (tl;dr: `GitSdin` was deleting the final newline which caused reading exactly `size` bytes to fail -- that's why I just deleted `GitStdin`, only this function was using it).

internal/reorder/plan.go
```diff
@@ -0,0 +1,71 @@
...
+		// Need to figure out the upstream commit to figure out the list of
+		// commits associated with this branch.
+		var upstreamCommit string
+		// TODO: would be nice to show the user whether or not the branch is
+		// 		already up-to-date with the parent.
+		if branch.Parent.Head != "" {
+			branchCmd.Parent = branch.Parent.Name
+			upstreamCommit = branch.Parent.Head
+		} else {
+			trunkCommit, err := repo.MergeBase(&git.MergeBase{
```
draftcode, line 31: Hm.

Wonder why this is not using `branch.Parent.Head`. Can you add a comment on this?

internal/reorder/plan_test.go
```diff
@@ -0,0 +1,80 @@
+package reorder
+
+import (
+	"github.com/aviator-co/av/internal/git"
+	"github.com/aviator-co/av/internal/git/gittest"
+	"github.com/aviator-co/av/internal/meta"
+	"github.com/aviator-co/av/internal/meta/jsonfiledb"
+	"github.com/stretchr/testify/require"
+	"testing"
```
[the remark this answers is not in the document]
travigd: Looks like the `goimports` pre-commit hook wasn't actually writing files. I'll fix that.

internal/reorder/plan.go
travigd: It's documented on `BranchState.Head`:
```
	// The commit SHA of the parent's latest commit. This is used when syncing
	// the branch with the parent to identify the commits that belong to the
	// child branch (since the HEAD of the parent branch may change).
	// This will be unset if Trunk is true.
```

A lot of that stuff shows its age from the very beginning of `av` though. `BranchState` should probably be like `ParentState` for clarity and `Trunk`/`Head` should be combined (since `Trunk` is `true` iff `Head` is empty -- which is annoying since the default is that `Trunk` is `false` and `Head` is empty which is an illegal state).

Generally it's just confusing wherever its used and seems to require keeping a lot of that info just in your head. I don't really want to add a comment here specifically because then we just end up copying the field documentation everywhere, but at some point I would like to change it to make it a bit more intuitive.

travigd: Also, to clarify, it's unset when `Trunk` is true because we can always derive it as `git merge-base <trunk> <branch>` -- we only need to store the parent head for non-root branches because the parent can change/be amended, at which point we can no longer tell what commits belong to the old version of the parent and the branch itself unless we do this.

draftcode: Sorry. Talked offline. I somehow read the condition the opposite way.

draftcode (approved):

PR #162, opened 2023-06-06
[MER-1962] Support retroactively rename the branches

draftcode:
When a user renames a branch with git, this allows them to retroactively
fit the Av metadata as well.

### DIFF
--- a/cmd/av/stack_branch.go
+++ b/cmd/av/stack_branch.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"emperror.dev/errors"
 	"github.com/aviator-co/av/internal/actions"
@@ -36,13 +37,8 @@ given as the first argument to the command. Branches should only be renamed
 with this command (not with git branch -m ...) because av needs to update
 internal tracking metadata that defines the order of branches within a stack.`,
 	SilenceUsage: true,
+	Args:         cobra.ExactArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) (reterr error) {
-		if len(args) != 1 {
-			_ = cmd.Usage()
-			return errors.New("exactly one branch name is required")
-		}
-		branchName := args[0]
-
 		repo, err := getRepo()
 		if err != nil {
 			return err
@@ -53,6 +49,7 @@ internal tracking metadata that defines the order of branches within a stack.`,
 			return err
 		}
 
+		branchName := args[0]
 		if stackBranchFlags.Rename {
 			return stackBranchMove(repo, db, branchName, stackBranchFlags.Force)
 		}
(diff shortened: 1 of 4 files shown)


cmd/av/stack_branch.go
```diff
@@ -159,6 +167,14 @@ func init() {
 	// See the comment on stackBranchFlags.Rename.
 	stackBranchCmd.Flags().
 		BoolVarP(&stackBranchFlags.Rename, "rename", "m", false, "rename the current branch")
+	stackBranchCmd.Flags().StringVar(
+		&stackBranchFlags.RetroactivelyRenameFrom,
+		"retroactively-rename-from",
```
travigd, line 172: I think this flag name is too long/clunky.

Can we just add a command like `av branch-meta set old-branch --name new-branch`?

draftcode: Yeah. This is a bit intentional because I don't think this is something people use daily basis, but need it as a way to fix this. We hide branch-meta because its more like a debugging command, so I don't want it documented.

cmd/av/stack_branch.go
```diff
@@ -36,13 +37,17 @@ given as the first argument to the command. Branches should only be renamed
...
 		if len(args) != 1 {
-			_ = cmd.Usage()
-			return errors.New("exactly one branch name is required")
+			return errors.Errorf("accepts 1 arg, received %d", len(args))
 		}
-		branchName := args[0]
-
+		c := strings.Count(args[0], ":")
+		if c > 1 {
+			return errors.New("the branch name should be NEW_BRANCH or OLD_BRANCH:NEW_BRANCH")
```
travigd, line 46: I think this check should just be inside the `stackBranchMove` function.

cmd/av/stack_branch.go
```diff
@@ -177,9 +183,17 @@ func stackBranchMove(
...
 	force bool,
 ) (reterr error) {
-	oldBranch, err := repo.CurrentBranchName()
-	if err != nil {
-		return err
+	var oldBranch string
+	if strings.ContainsRune(newBranch, ':') {
+		ss := strings.Split(newBranch, ":")
+		oldBranch = ss[0]
+		newBranch = ss[1]
```
travigd, line 190:
```suggestion
		oldBranch, newBranch, _ = strings.Cut(newBranch, ":")
```

cmd/av/stack_branch.go
```diff
@@ -236,11 +250,15 @@ func stackBranchMove(
...
-		return errors.WrapIff(err, "failed to rename Git branch")
+	if ok, err := repo.DoesBranchExist(oldBranch); err != nil {
+		return err
+	} else if ok {
+		if _, err := repo.Run(&git.RunOpts{
+			Args:      []string{"branch", "-m", newBranch},
+			ExitError: true,
+		}); err != nil {
+			return errors.WrapIff(err, "failed to rename Git branch")
+		}
```
travigd, line 261: Let's add an `else` clause here that prints out a message to the effect of

```
Branch <oldbranch> does not exist in Git. Updating av internal metadata only.
```

draftcode: Done

cmd/av/stack_branch.go
draftcode: Done.

cmd/av/stack_branch.go
draftcode: Done.

travigd (approved):

PR #164, opened 2023-06-06
docs: Add manpage for `av stack diff`

ErichKramer:

### DIFF
--- a/docs/av-stack-diff.1.md
+++ b/docs/av-stack-diff.1.md
@@ -1,0 +1,13 @@
+# av-stack-diff 1 "" av-cli "Aviator CLI User Manual"
+
+# NAME
+
+av-stack-diff - Show the diff between working tree and parent branch. 
+
+# SYNOPSIS
+
+`av stack diff`
+
+# DESCRIPTION
+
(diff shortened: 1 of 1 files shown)


docs/av-stack-diff.1.md
ErichKramer, line 1: Can you tell that I didn't use `av` for these commits

docs/av-stack-diff.1.md
```diff
@@ -0,0 +1,17 @@
+# av-stack-diff 1 "" av-cli "Aviator CLI User Manual"
+
+# NAME
+
+av-stack-diff - Show the diff between working tree and parent branch. 
+
+# SYNOPSIS
+
+`av stack diff [-h | --help]`
```
travigd, line 9: We haven't been including `--help` in the other commands I don't think, so let's not include it here for consistency. You can just delete the `OPTIONS` section altogether since there are no other flags.

ErichKramer: done

docs/av-stack-diff.1.md
ErichKramer: Well, now I've rewritten history.

PR #165, opened 2023-06-06
[MER-2418] Make it easy to detect a PR markdown

draftcode:
This is a nice to have and a harmless change. The temp file created for
the pull-request title and body now has an extension '.av.md'. This is
anyway recognized as a markdown file for most of the editors, but this
extra 'av' extension allows us to have a special file type that supports
'%%' as a comment.

As an example, added a Vim plugin that recognize this file type and
treats '%%' as a comment.

### DIFF
--- a/internal/actions/pr.go
+++ b/internal/actions/pr.go
@@ -284,7 +284,7 @@ func CreatePullRequest(
 
 		res, err := editor.Launch(repo, editor.Config{
 			Text:           editorText,
-			TmpFilePattern: "pr-*.md",
+			TmpFilePattern: "pr-*.av.md",
 			CommentPrefix:  "%%",
 		})
 		if err != nil {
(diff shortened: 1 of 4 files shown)


draftcode:
This makes Vim render the PR body like this:

image

draftcode:
I'm fine with excluding the Vim plugin part. I can keep it local. But the PR temp file change is anyway harmless.

jainankit (approved):
This is great. We should also document this - suggest that folks can use the vim plugin or build their own for their IDE (i future we can do a VS code plugin for this).

travigd:
I wonder if it's possible to include these as extra files installed by (e.g.,) the brew cask.

draftcode:
> I wonder if it's possible to include these as extra files installed by (e.g.,) the brew cask.

It's probably a bit difficult. But since Vim has relatively established plugin install system now, I think it's easy to distribute this as a Vim plugin (by moving these files to another repo).

PR #168, opened 2023-06-06
refac:[MER-2419] Split pr_create.go out of pr.go

ErichKramer:


ErichKramer:
hmm. Will putting the linear ticket update my actual underlying ticket before I want it to? Hm

PR #173, opened 2023-06-07
feat: add av pr status

doratzeng:
I'll add BotPR data in my next PR

passing checks - 
Screen Shot 2023-06-07 at 9 54 36 PM

pending checks - 
Screen Shot 2023-06-07 at 9 59 48 PM

merged pr - 
Screen Shot 2023-06-07 at 9 56 25 PM

### DIFF
--- a/cmd/av/pr_status.go
+++ b/cmd/av/pr_status.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/aviator-co/av/internal/avgql"
 	"github.com/aviator-co/av/internal/utils/colors"
@@ -39,6 +40,7 @@ var query struct {
 					Conclusion graphql.String
 				}
 			}
+			// TODO: add BotPR info
 		} `graphql:"pullRequest(number: $prNumber)"`
 	} `graphql:"githubRepository(owner: $repoOwner, name:$repoName)"`
 }
@@ -67,7 +69,9 @@ var prStatusCmd = &cobra.Command{
 		// Print PR info
 		_, _ = fmt.Fprint(
 			os.Stderr,
-			"#162 ",
+			"#",
+			variables["prNumber"],
+			" ",
 			colors.UserInput(query.GithubRepository.PullRequest.Title),
 			"\n",
 		)
@@ -94,7 +98,7 @@ var prStatusCmd = &cobra.Command{
 			os.Stderr,
 			indent,
 			"Created at: ",
-			colors.UserInput(query.GithubRepository.PullRequest.CreatedAt),
+			colors.UserInput(formatTimestamp(string(query.GithubRepository.PullRequest.CreatedAt))),
 			"\n",
 		)
 
@@ -103,7 +107,9 @@ var prStatusCmd = &cobra.Command{
 				os.Stderr,
 				indent,
 				"Queued at: ",
-				colors.UserInput(query.GithubRepository.PullRequest.QueuedAt),
+				colors.UserInput(
+					formatTimestamp(string(query.GithubRepository.PullRequest.QueuedAt)),
+				),
 				"\n",
 			)
 		}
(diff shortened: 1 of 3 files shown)


cmd/av/pr_status.go
```diff
@@ -0,0 +1,218 @@
...
+	if result == "SUCCESS" {
+		return "\u2705"
+	} else if result == "FAILURE" {
+		return "\u274C"
+	} else {
+		return "\u231B"
+	}
+}
+
+func formatTimestamp(timeString string) string {
```
travigd, line 207: We should probably have a package that just contains utilities like this. Maybe like `utils/timeutils` so that we have `timeutils.FormatLocal(t time.Time) string`.

cmd/av/pr_status.go
```diff
@@ -0,0 +1,218 @@
...
+	var variables = map[string]interface{}{
+		"repoOwner": graphql.String(repository.Owner),
+		"repoName":  graphql.String(repository.Name),
+		"prNumber":  graphql.Int(prNumber),
+	}
+	return variables, nil
+}
+
+func emojiForRequiredCheckResult(result string) string {
+	if result == "SUCCESS" {
```
travigd, line 198: Use `switch` rather than `if` here

cmd/av/pr_status.go
```diff
@@ -0,0 +1,218 @@
...
+	if !exists {
+		return nil, errors.New("could not find current branch")
+	}
+
+	if branch.PullRequest == nil {
+		return nil, errors.New("pull request does not exist")
+	}
+
+	prNumber := branch.PullRequest.Number
+	repository, _ := tx.Repository()
```
travigd, line 188: We actually do need to check if this is set, because if the repo owner/name is empty, this query will fail.

cmd/av/pr_status.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	tx := db.ReadTx()
+
+	currentBranchName, err := repo.CurrentBranchName()
+	if err != nil {
+		return nil, err
+	}
+
+	branch, exists := tx.Branch(currentBranchName)
```
travigd, line 178: In this place, you can actually skip this `exists` check since we'll get to the `PullRequest` check below.

cmd/av/pr_status.go
```diff
@@ -0,0 +1,218 @@
...
+	if !exists {
+		return nil, errors.New("could not find current branch")
+	}
+
+	if branch.PullRequest == nil {
+		return nil, errors.New("pull request does not exist")
```
travigd, line 184: Error messages are super important in this kind of context. This one isn't actionable. A better one might be (e.g.,) `` this branch has no associated pull request (run `av pr create` to create one) ``.

travigd (commented):
Left a few comments. Let me know if you disagree or wanna chat about thing! 😸

internal/utils/timeutils/timeutils.go
```diff
@@ -0,0 +1,20 @@
...
+	outputLayout := "2 January 2006 3:04:05 PM PST"
+
+	timestamp, err := time.Parse(inputLayout, timeString)
+	if err != nil {
+		return timeString
+	}
+
+	timestampDefaultTZ := timestamp.Local()
+	return timestampDefaultTZ.Format(outputLayout)
+}
```
draftcode, line 20: The input layout should use `time.RFC3339Nano` constant instead. Also this should be parsed on the caller side. Is there a way to make the graphql library parse the time.Time by defining a literal type or something?

doratzeng: the library doesnt parse datetime :( they have an open issue, i don't think it's really being worked on https://github.com/shurcooL/githubv4/issues/9

draftcode: Looks like it works https://github.com/shurcooL/githubv4#scalar-types

travigd (approved):

draftcode (approved):

PR #174, opened 2023-06-08
feat(stack reorder): Delete branches removed from reorder plan

travigd:
# Example

Have a branch structure:
```
* 2023-06-08-two-b ce20b00  (HEAD -> 2023-06-08-two)
| 
* 2023-06-08-two-a 1f8f16b 
| 
* 2023-06-08-one-b 061ed6e  (2023-06-08-one)
| 
* 2023-06-08-one-a a4db90d 
| 
* 2023-05-30-one-a (#957) 87232f0  (origin/main, origin/HEAD, main)
```

Run `av stack reorder` and delete the `stack-branch 2023-06-08-two` line (effectively merging all commits into `2023-06-08-one`):

image

Choose `e` to edit the plan:
image
(looks like unfortunately we lose the comment information, but this can be fixed in a future change)

Save the same plan, choose `o` to orphan the branches:
image

Confirmed that `2023-06-08-two` still exists and hasn't had it's HEAD modified.

Reset to the state above, but this time choose `d` to delete the branches:
image

### DIFF
--- a/internal/reorder/deletebranch.go
+++ b/internal/reorder/deletebranch.go
@@ -1,0 +1,61 @@
+package reorder
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/aviator-co/av/internal/git"
+	"github.com/aviator-co/av/internal/utils/colors"
+	"github.com/aviator-co/av/internal/utils/errutils"
+)
+
+type DeleteBranchMode string
+
+// DeleteBranchCmd is a command that deletes a branch.
+type DeleteBranchCmd struct {
+	Name string
+	// If true, delete the branch from Git as well as from the internal database.
+	// If false, only delete the branch metadata from the internal database.
+	DeleteRef bool
+}
+
+func (d DeleteBranchCmd) Execute(ctx *Context) error {
+	tx := ctx.DB.WriteTx()
+	tx.DeleteBranch(d.Name)
+	if err := tx.Commit(); err != nil {
+		return err
+	}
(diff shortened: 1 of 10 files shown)


internal/reorder/deletebranch.go
```diff
@@ -0,0 +1,93 @@
...
+		"Deleted branch ", colors.UserInput(d.Name), ".\n",
+	)
+	return nil
+}
+
+func (d DeleteBranchCmd) String() string {
+	sb := strings.Builder{}
+	sb.WriteString("delete-branch ")
+	sb.WriteString(d.Name)
+	if d.DeleteRef {
```
draftcode, line 65: --delete-ref looks confusing? A branch is a ref.

Aside from that, do we want to have this command? It seems that simply deleting a line would be suffice to support deleting branches, and for I cannot think a good use case for this command.

travigd: Yeah, I meant it to disambiguate deleting av's version of the branch and deleting the actual Git ref. It could also be `--delete-git` but I'm not sure if that's actually any clearer. Maybe `--delete-git-ref` would be the clearest of all.

But to your point, the only reason I implemented this as a command is convenience of implementation. It's nice to be able to just append the `DeleteBranch` commands to the reorder plan if we detect that the branches have been removed from the plan (which is what we do here). We probably don't even need to document that this exists. We only even need the flag version because the command is serialized as a string whenever we write the continuation to file.

draftcode (approved):

PR #175, opened 2023-06-08
feat: add associated botpr info to av pr status

doratzeng:

### DIFF
--- a/cmd/av/pr_status.go
+++ b/cmd/av/pr_status.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/aviator-co/av/internal/avgql"
 	"github.com/aviator-co/av/internal/utils/colors"
@@ -39,6 +40,7 @@ var query struct {
 					Conclusion graphql.String
 				}
 			}
+			// TODO: add BotPR info
 		} `graphql:"pullRequest(number: $prNumber)"`
 	} `graphql:"githubRepository(owner: $repoOwner, name:$repoName)"`
 }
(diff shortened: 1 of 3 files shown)


cmd/av/pr_status.go
```diff
@@ -150,6 +159,28 @@ var prStatusCmd = &cobra.Command{
...
 		}
 
+		// Get Associated BotPR info
+		botPullRequest := query.GithubRepository.PullRequest.BotPullRequest
+		if botPullRequest.Number == 0 {
+			// no associated botPR
+			return nil
+		}
+
+		_, _ = fmt.Fprint(os.Stderr, "Associated PR #", botPullRequest.Number, " Required Checks\n")
```
travigd, line 169: Can we call it bot pull request? Let's try to use consistent language. It's called "associated PR" in a few places, but we should either change that to "bot pull request" or change what we call it in the API so those two things match.

cmd/av/pr_status.go
```diff
@@ -150,6 +159,28 @@ var prStatusCmd = &cobra.Command{
...
+
+		botRequiredCheckStatuses := query.GithubRepository.PullRequest.BotPullRequest.RequiredCheckStatuses
+		for index := range botRequiredCheckStatuses {
```
travigd, line 172:
```suggestion
		for _, status := range botRequiredCheckStatuses {
```

Doing `for i := range slice` yields the indices, but you can also do `for i, elt := range slice` to get the indices and their values (and if you don't need the value, you can replace the `i` with a `_` to discard it). Same works for iterating over maps (first form iterates over keys, second form iterates over keys and values).

travigd (approved):

PR #176, opened 2023-06-08
[MER-2446] Adjust styles for web and manpages

draftcode:
We are trying to use `docs/*` as both command manuals on the web as well
as online manual distributed along with the binary. For that, we want to
customize go-md2man output, especially the header part.

This includes a fork of https://github.com/cpuguy83/go-md2man. This
includes the title part and some whitespace handling
modifications.

Note that go-md2man is licensed under MIT license, which is the same as
this repository. The file independently have a license header and a
copyright notice with a SPDX header.

### DIFF
--- a/cmd/av/pr.go
+++ b/cmd/av/pr.go
@@ -5,13 +5,12 @@ import (
 )
 
 var prCmd = &cobra.Command{
-	Use: "pr",
+	Use:   "pr",
 	Short: "manage pull requests",
 }
 
 func init() {
 	prCmd.AddCommand(
 		prCreateCmd,
-
 	)
 }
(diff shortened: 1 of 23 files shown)


internal/md2man/md2man.go
```diff
@@ -0,0 +1,20 @@
+package md2man
```
travigd, line 1: This isn't a super strong opinion, but could we put this underneath `docs/internal` or something? I don't expect we'd use this as part of `av` itself.[](url)

travigd (approved):

internal/md2man/md2man.go
draftcode: Done.

PR #177, opened 2023-06-08
feat: add force flag to stack branch rename

doratzeng:

### DIFF
--- a/cmd/av/stack_branch.go
+++ b/cmd/av/stack_branch.go
@@ -1,10 +1,15 @@
 package main
 
 import (
+	"fmt"
+	"os"
+
 	"emperror.dev/errors"
+	"github.com/aviator-co/av/internal/actions"
 	"github.com/aviator-co/av/internal/git"
 	"github.com/aviator-co/av/internal/meta"
 	"github.com/aviator-co/av/internal/utils/cleanup"
+	"github.com/aviator-co/av/internal/utils/colors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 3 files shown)


cmd/av/stack_branch.go
```diff
@@ -196,6 +202,14 @@ func stackBranchMove(
 	currentMeta.Name = newBranch
 	tx.SetBranch(currentMeta)
 
+	if !force {
+		if currentMeta.PullRequest != nil {
+			return errors.New(
+				"cannot rename branch because a pull request already exists (bypass with the `--force` flag, but this renames your local branch and not its remote counterpart)",
```
jainankit, line 208:
```suggestion
				"cannot rename branch because a pull request already exists (bypass with the `--force` flag, but this will require closing this pull request and opening a new one using `av pr create`)",
```

jainankit (approved):

cmd/av/stack_branch.go
```diff
@@ -193,9 +199,18 @@ func stackBranchMove(
 			Trunk: true,
 		}
 	}
+	currentMeta.PullRequest = nil
 	currentMeta.Name = newBranch
 	tx.SetBranch(currentMeta)
 
+	if !force {
+		if currentMeta.PullRequest != nil {
```
travigd, line 207: Looks like you always set `PullRequest = nil` before this check.

cmd/av/stack_branch.go
```diff
@@ -193,9 +199,18 @@ func stackBranchMove(
...
+			return errors.New(
+				"cannot rename branch because a pull request already exists (bypass with the `--force` flag, but this will require closing this pull request and opening a new one using `av pr create`)",
```
travigd, line 209: This is a very long error message. It might make more sense to just print some output to stderr and return a shorter error message.
```go
_, _ = fmt.Fprint(os.Stderr,
    colors.Failure("Cannot rename branch <branchname>: pull request #<number> would be orphaned.\n"),
    colors.Faint("  - Use --force to override this check."),
)

return actions.ErrExitSilently{ExitCode: 127}
```

PR #186, opened 2023-06-29
[MER-2335] Add a manpage for av stack reorder

draftcode:
This is a bare minimum. We can add more explaination later.

### DIFF
--- a/docs/av-stack-reorder.1.md
+++ b/docs/av-stack-reorder.1.md
@@ -1,0 +1,27 @@
+# av-stack-reorder
+
+## NAME
+
+av-stack-reorder - Interactively reorder the stack
+
+## SYNOPSIS
+
+```synopsis
+av stack reorder [--continue | --abort]
+```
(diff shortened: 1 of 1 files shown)


docs/av-stack-reorder.1.md
```diff
@@ -0,0 +1,27 @@
...
+av stack reorder [--continue | --abort]
+```
+
+## DESCRIPTION
+
+This is analogous to git rebase --interactive but operates across all branches
+in the stack.
+
+Branches can be re-arranged within the stack and commits can be edited,
+squashed, dropped, or moved within the stack.
```
jainankit, line 19: Might be worth explicitly saying that the commits can be moved across the branches.

jainankit (approved):

docs/av-stack-reorder.1.md
draftcode: Done.

PR #188, opened 2023-07-11
[MER-2419] feat: `av pr queue` support

ErichKramer:


cmd/av/pr_queue.go
```diff
@@ -0,0 +1,110 @@
...
+		fmt.Fprintf(
+			os.Stdout, 
+			string(pr),
+			"\n",
+		)
+		return nil
+	},
+}
+
+func getMutationVariables() (map[string]interface{}, error) {
```
travigd, line 61: this function is in the `main` namespace so pls name it more descriptively. I'd honestly prefer just inlining this into `prQueueCmd.RunE` but I'm weird that way.

cmd/av/pr_queue.go
```diff
@@ -0,0 +1,110 @@
...
+		// I have a feeling this would be better written inside of av/internals
+		client := avgql.NewClient()
+
+		var mutation struct {
+			QueuePullRequest struct {
+				UpdateMergeQueueStatusPayload struct {
+					PullRequest	graphql.String
+				}
+					
+			} `graphql:"githubRepository(input: {owner: $repoOwner, name:$repoName})"`
```
travigd, line 42: you can just pass the `input` as a `map[string]any` fyi, but either way works.

```go
variables := map[string]any{
  "input": map[string]any{
    "owner": ...,
    "name": ...,
  },
}
```

travigd: wait is this the right annotation here? I think would generate a mutation that looks like

```
mutation (...) {
  githubRepository(input: {...}) {...}
}
```
rather than
```
mutation (...) {
  queuePullRequest(input: {...}) {...}
}
```

cmd/av/pr_queue.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		err = client.Mutate(context.Background(), &mutation, variables)
+		if err != nil {
+			return err
+		}
+		// We print to stderr everywhere, why?
```
travigd, line 48: We print to stderr because we don't want users to rely on the output of `stdout`. imo all logging should go to stderr and stdout should only be used for programatic output.

ErichKramer: Done

cmd/av/pr_queue.go
ErichKramer: Fixed

cmd/av/pr_queue.go
ErichKramer: Disagree on the in-lining, I think pulling this out improves readability.

This pattern is borrowed from `pr_status.go`, open to function renaming recommendations?

travigd: What do you think about moving the actual mutation into this function then, and just returning the pull request info? Right now, this function is separate from the `var mutation` definition, but also it's very coupled to it (the keys in the map here have to match the variables used in the mutation). It's always a yellow/red-ish flag when you have that tight of coupling.

That being said, it's still within the same file, just ~20 lines away, so it's not the end of the world. If you just want to rename this function, we could just do `getQueuePullRequestMutationVariables`. It's obviously long but it's only used once so 🤷.

cmd/av/pr_queue.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+
+
+		// I have a feeling this would be better written inside of av/internals
+		client := avgql.NewClient()
+
+		var mutation struct {
+			QueuePullRequest struct {
+				QueuePullRequestPayload struct {
+					PullRequest	graphql.String
```
travigd, line 39: I think we need to actually finish the backend before merging this, because converting the `pullRequest` field from string to an actual GraphQL object would be a breaking change.

cmd/av/pr_queue.go
```diff
@@ -0,0 +1,110 @@
...
+		"repoName":  graphql.String(repository.Name),
+		"prNumber":  graphql.Int(prNumber),
+	}
+	return variables, nil
+}
+
+func init() {
+	prQueueCmd.Flags().BoolVar(
+		&prQueueFlags.SkipLine, "skip-line", false,
+		"skip in front of the existing PRs, merge this PR right now",
```
travigd, line 104: nit: let's always use "pull request" rather than "PR" in user-facing interfaces

```suggestion
		"queue this pull request ahead of all others in the queue",
```

cmd/av/pr_queue.go
```diff
@@ -0,0 +1,110 @@
...
+	)
+	prQueueCmd.Flags().StringVarP(
+		&prQueueFlags.Targets, "targets", "t", "",
+		"additional targets affected by this PR",
```
travigd, line 108: same here re: `pull request`

cmd/av/pr_queue.go
ErichKramer: Got it

ErichKramer: Im grinding on this, bitten by circular dependencies. I think in the sake of shipping the best move is to simply breakup queue.py on the backend.

ErichKramer: Done; see: https://github.com/aviator-co/mergeit/pull/3053

cmd/av/pr_queue.go
ErichKramer: Done

cmd/av/pr_queue.go
ErichKramer: Done

cmd/av/pr_queue.go
ErichKramer: I in-lined it.

ErichKramer: done

cmd/av/pr_queue.go
```diff
@@ -0,0 +1,103 @@
...
+	SkipLine bool
+	Targets  string
+}
+
+var prQueueCmd = &cobra.Command{
+	Use:          "queue",
+	Short:        "queue a pull request for the current branch",
+	SilenceUsage: true,
+	Args:         cobra.NoArgs,
+	// error or reterr from emperror.dev/errors here?
```
travigd, line 28: Doesn't reeeeally matter that much. The main benefit is that if you run with `av --debug` it will print out stack traces, but the traces would be so shallow at this point that it doesn't matter super hard.

cmd/av/pr_queue.go
```diff
@@ -0,0 +1,103 @@
...
+			return actions.ErrRepoNotInitialized
+		}
+
+		var variables = map[string]interface{}{
+			"repoOwner": graphql.String(repository.Owner),
+			"repoName":  graphql.String(repository.Name),
+			"prNumber":  graphql.Int(prNumber),
+		}
+
+		// I have a feeling this would be better written inside of av/internals
```
travigd, line 65: Remove this comment~ we can move things to `internal` at some point if we want to.

travigd (approved):

PR #194, opened 2023-07-26
feat: Allow specifying Aviator API URL from configuration

travigd:
Feature request from a customer to support on-prem installations.

Ideally this would be configurable per-repo as it's possible that a user might have some repositories that use the \~\~cloud~~ version of Aviator and some that use on-prem, but that's not very likely at this point and I didn't want to add that complexity to this pull request.

### DIFF
--- a/internal/avgql/client.go
+++ b/internal/avgql/client.go
@@ -2,21 +2,29 @@ package avgql
 
 import (
 	"context"
+	"net/url"
 	"os"
 
+	"emperror.dev/errors"
 	"github.com/aviator-co/av/internal/config"
 	"github.com/shurcooL/graphql"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/oauth2"
 )
 
-func NewClient() *graphql.Client {
+func NewClient() (*graphql.Client, error) {
 	src := oauth2.StaticTokenSource(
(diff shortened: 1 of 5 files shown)


internal/avgql/client.go
```diff
@@ -2,21 +2,29 @@ package avgql
...
-func NewClient() *graphql.Client {
+func NewClient() (*graphql.Client, error) {
 	src := oauth2.StaticTokenSource(
 		&oauth2.Token{AccessToken: config.Av.Aviator.APIToken},
 	)
 	httpClient := oauth2.NewClient(context.Background(), src)
 	apiURL := os.Getenv("AV_GRAPHQL_URL")
 	if apiURL == "" {
-		apiURL = "https://api.aviator.co/graphql"
+		baseURL, err := url.Parse(config.Av.Aviator.APIHost)
```
ErichKramer, line 22: should config take priority or should env var take priority?

Config taking prio makes sense to me.

draftcode (approved):

ErichKramer (approved):

jainankit (approved):

internal/avgql/client.go
travigd: I think environment variables should usually take priority since it's easier to override them in more local contexts. My order is always:
* Defaults
* Config files
* Environment variables
* Flags

PR #198, opened 2023-08-21
[MER-2188] Use gh auth token if it's available

draftcode:

### DIFF
--- a/cmd/av/main.go
+++ b/cmd/av/main.go
@@ -1,8 +1,11 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"os"
+	"os/exec"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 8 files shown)


cmd/av/main.go
```diff
@@ -149,6 +152,22 @@ func checkCliVersion() {
 var once sync.Once
 var lazyGithubClient *gh.Client
 
+func discoverGitHubAPIToken() string {
```
travigd, line 155: Probably just makes sense to call this as part of `getClient` and return the "github token must be set" error from `getClient`.

travigd (approved):
Looks good. I think @ErichKramer looked into this at some point and identified some reason it might not actually work, so we should get his input on this.

cmd/av/main.go
draftcode: Done.

PR #199, opened 2023-08-24
[MER-2814] Automatically make a PR draft

draftcode:
This creates a PR in a draft state when the title contains "WIP". Can be
turned off in the config.

### DIFF
--- a/cmd/av/pr_create.go
+++ b/cmd/av/pr_create.go
@@ -69,6 +69,11 @@ Examples:
 			prCreateFlags.Body = string(bodyBytes)
 		}
 
+		draft := config.Av.PullRequest.Draft
+		if cmd.Flags().Changed("draft") {
+			draft = prCreateFlags.Draft
+		}
+
 		if _, err := actions.CreatePullRequest(
 			context.Background(), repo, client, tx,
 			actions.CreatePullRequestOpts{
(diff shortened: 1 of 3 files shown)


travigd (commented):
I have mixed feelings on this. It's, to some extent, feature bloat and also a feature that might end up being surprising to people who don't know about it.

The use case for this is presumably just forgetting to pass `--draft` on the command line when running `av pr create` and I don't see a strong reason to justify this when passing `--draft` is still pretty easy. (I'm generally in favor of trying to have a forgiving UX that is build with an awareness that people make mistakes and forget to do things, but I don't know if the tradeoff is worth it here).

I've imagined before having a more structured way to do this kind of thing. I wonder if we could so something like have special "comment directives" in the editor that can change the behavior. For example, maybe adding a `%%draft` line in the editor, or having something like [Markdown frontmatter](https://daily-dev-tips.com/posts/what-exactly-is-frontmatter/). That way it's also discoverable -- we can include instructions at the bottom of the editor (where we currently list all the commits that will be included as part of the pull request).

All that being said, this is a pretty small change, so if you feel strongly, more than happy to just approve. Let me know your thoughts!

draftcode:
I found myself forgetting to set --draft after the PR is created multiple times, so if this is an acceptable feature, then I'd like to proceed with this.

doratzeng (approved):

PR #200, opened 2023-08-29
[MER-2744] fix: look for correct git dir path.

ErichKramer:

### DIFF
--- a/cmd/av/helpers.go
+++ b/cmd/av/helpers.go
@@ -18,18 +18,25 @@ var cachedRepo *git.Repo
 
 func getRepo() (*git.Repo, error) {
 	if cachedRepo == nil {
-		cmd := exec.Command("git", "rev-parse", "--show-toplevel")
+		cmd := exec.Command("git", "rev-parse", "--path-format=absolute", "--show-toplevel", "--git-common-dir")
+
 		if rootFlags.Directory != "" {
 			cmd.Dir = rootFlags.Directory
 		}
-		toplevel, err := cmd.Output()
+		paths, err := cmd.Output()
 		if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/av/init.go
```diff
@@ -52,7 +52,7 @@ var initCmd = &cobra.Command{
 
 		cu.Cancel()
 		if err := tx.Commit(); err != nil {
-			return err
+			return fmt.Errorf("Could not get write to local DB: %w", err)
```
travigd, line 55: No need to wrap these errors. Generally we only wrap where we can provide additional context, but in these cases, we're not providing any additional context (the reason for the failure to write should be provided by the error returned by `tx.Commit`), so don't wrap them.

travigd: I think all of the error wraps you've added in this file are unnecessary, can you just remove them all?

cmd/av/helpers.go
```diff
@@ -18,18 +18,21 @@ var cachedRepo *git.Repo
...
 		if rootFlags.Directory != "" {
 			cmd.Dir = rootFlags.Directory
 		}
-		toplevel, err := cmd.Output()
+		gitPath, err := cmd.Output()
 		if err != nil {
 			return nil, errors.Wrap(
 				err,
-				"failed to determine repo toplevel (are you running inside a Git repo?)",
+				"failed to determine current worktree's git directory (are you running inside a Git repo?)",
```
travigd, line 30: Most users won't know what a worktree is, let's try to make this error a little bit more friendly.
```suggestion
				"failed to find .git directory (are you running inside a Git repo?)",
```

draftcode (approved):

PR #213, opened 2023-11-20
feat: Add `--reviewers` to `av pr create`

travigd:

### DIFF
--- a/cmd/av/pr_create.go
+++ b/cmd/av/pr_create.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"io"
 	"os"
-	"strings"
 
 	"emperror.dev/errors"
 	"github.com/aviator-co/av/internal/actions"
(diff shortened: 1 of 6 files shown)


cmd/av/pr_create.go
```diff
@@ -85,12 +88,22 @@ Examples:
...
+		if err != nil {
 			return err
 		}
 		if err := tx.Commit(); err != nil {
 			return err
 		}
+
+		// Do this after creating the PR and committing the transaction so that
+		// our local database is up-to-date even if this fails.
+		if len(prCreateFlags.Reviewers) > 0 {
```
jainankit, line 101: Doesn't GH allow passing reviewers as part of creating the PR? I wonder if it does, would the behavior be different to do this in two separate steps. Would GH assign some default reviewers still

internal/actions/reviewers.go
```diff
@@ -0,0 +1,74 @@
...
+// It accepts a list of reviewers, which can be either GitHub user logins or
+// team names in the format `@organization/team`.
+func AddPullRequestReviewers(
+	ctx context.Context,
+	client *gh.Client,
+	prID githubv4.ID,
+	reviewers []string,
+) error {
+	_, _ = fmt.Fprint(os.Stderr,
+		"  - adding ", colors.UserInput(len(reviewers)), " reviewers to pull request\n",
```
jainankit, line 25:
```suggestion
		"  - adding ", colors.UserInput(len(reviewers)), " reviewer(s) to pull request\n",
```

cmd/av/pr_create.go
travigd: > Doesn't GH allow passing reviewers as part of creating the PR?

The create pull request API (mutation `createPullRequest`) does not have a way to assign reviewers (nor does the `updatePullRequest` mutation).

jainankit (approved):

PR #218, opened 2023-12-20
[MER-3309] Add an extra merge commit detection based on commit messages

draftcode:
When a PR gets closed by "Close #123", usually GitHub marks the PR to be
closed by certain commit. However, this mechanism is not reliable, and
sometimes GitHub doesn't mark it so and just close it. This breaks the
merge commit detection since av-cli doesn't have a way to know it from
GitHub API.

By using git-log, we can figure out whether there's actually a commit
that closes the PRs with this comment. This adds that extra detection
method.

In order to make this work, it moves the timing of git-fetch. This is
needed in order for a local repository to have commits in the upstream
to check the commit messages.

### DIFF
--- a/internal/actions/sync_branch.go
+++ b/internal/actions/sync_branch.go
@@ -61,6 +61,10 @@ func SyncBranch(
 		}
 	} else {
 		if opts.Fetch {
+			fetchHead, err := fetchRemoteTrunkHead(repo, tx, branch)
+			if err != nil {
+				return nil, err
+			}
 			update, err := UpdatePullRequestState(ctx, client, tx, branch.Name)
 			if err != nil {
 				_, _ = fmt.Fprint(os.Stderr, colors.Failure("      - error: ", err.Error()), "\n")
@@ -77,6 +81,14 @@ func SyncBranch(
 						" (create one with ", colors.CliCmd("av pr create"),
 					" or ", colors.CliCmd("av stack submit"), ")\n",
 				)
+			} else if branch.PullRequest.State == githubv4.PullRequestStateClosed && branch.MergeCommit == "" {
+				branch.MergeCommit, err = findMergeCommitWithGitLog(repo, fetchHead, branch)
+				if err != nil {
+					return nil, errors.Wrap(err, "failed to find the merge commit from git-log")
+				}
+				if branch.MergeCommit != "" {
+					tx.SetBranch(branch)
+				}
 			}
 		}
 
@@ -243,15 +255,6 @@ func syncBranchRebase(
 			"  - rebasing ", colors.UserInput(branch.Name),
 			" on top of merge commit ", colors.UserInput(short), "\n",
 		)
-		if opts.Fetch {
-			if _, err := repo.Git("fetch", "origin", origParentBranch.MergeCommit); err != nil {
-				return nil, errors.WrapIff(
-					err,
-					"failed to fetch merge commit %q from origin",
-					short,
-				)
-			}
-		}
 
 		// Replay the commits from this branch directly onto the merge commit.
 		// The HEAD of trunk might have moved forward since this, but this is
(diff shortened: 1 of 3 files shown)


internal/actions/sync_branch.go
```diff
@@ -420,6 +423,36 @@ func syncBranchRebase(
 	return nil, nil
 }
 
+func fetchUpstreamCommits(repo *git.Repo, tx meta.WriteTx, branch meta.Branch) (string, error) {
```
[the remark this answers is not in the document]
travigd: Actually, maybe `fetchRemoteTrunkHead` would be a better name (it's my understanding that that's what this function is doing -- let me know if I misunderstood).

internal/actions/sync_branch.go
```diff
@@ -420,6 +423,36 @@ func syncBranchRebase(
...
+	parent := branch
+	for parent.Parent.Name != "" {
+		parent, _ = tx.Branch(parent.Parent.Name)
+	}
```
travigd, line 430: I think this could just be simplified to something like

```
trunk, ok := meta.Trunk(branch)
```

internal/git/log.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+var (
+	closeCommitPattern = regexp.MustCompile(`(?i)\b(?:close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved)\W+#(\d+)\b`)
+)
+
+type LogOpts struct {
+	// RevisionRange is the range of the commits specified by the format described in
+	// git-log(1).
+	RevisionRange string
```
travigd, line 21: Technically this should be `[]string`.

internal/git/log.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// Log returns a list of commits specified by the range.
+func (r *Repo) Log(opts LogOpts) ([]*CommitInfo, error) {
+	res, err := r.Run(&RunOpts{
+		Args:      []string{"log", "--format=%H%x00%h%x00%s%x00%b%x00", opts.RevisionRange},
```
travigd, line 27: Can you add a trailing `--` here to avoid the possibility of an ambiguous revision/path error? According to manpages, `git log` can take paths as well.

internal/actions/sync_branch.go
```diff
@@ -420,6 +423,36 @@ func syncBranchRebase(
...
+	return commitHash, nil
+}
+
+// findMergeCommitWithGitLog looks for the merge commit for a specified PR.
+//
+// Usually, GitHub should set which commit closes a pull request. This is known to be not that
+// reliable. When we cannot find a merge commit for a closed PR, we try to find if any commit in the
+// upstream closes the pull request as a fallback.
+func findMergeCommitWithGitLog(repo *git.Repo, upstreamCommit string, branch meta.Branch) (string, error) {
+	cis, err := repo.Log(git.LogOpts{RevisionRange: branch.Name + ".." + upstreamCommit})
```
travigd, line 448: Can we use the more explicit syntax here rather than the double dot? I don't always remember what the double dot means. Also would be helpful either way to have a comment explaining what we're querying.
```suggestion
    // Find all commits that have been merged into the trunk since this branch
	cis, err := repo.Log(git.LogOpts{RevisionRange: []string{upstreamCommit, "^" + branch.Name}})
```

(^assuming you incorporate my suggested change below about `RevisionRange` being a slice)

internal/actions/sync_branch.go
```diff
@@ -420,6 +423,36 @@ func syncBranchRebase(
...
+func fetchUpstreamCommits(repo *git.Repo, tx meta.WriteTx, branch meta.Branch) (string, error) {
+	parent := branch
+	for parent.Parent.Name != "" {
+		parent, _ = tx.Branch(parent.Parent.Name)
+	}
+
+	if _, err := repo.Git("fetch", "origin", parent.Name); err != nil {
+		return "", errors.WrapIff(err, "failed to fetch %q from origin", parent.Name)
+	}
+	commitHash, err := repo.RevParse(&git.RevParse{Rev: "FETCH_HEAD"})
```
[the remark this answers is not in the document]
draftcode: Depending on the fetch spec, the name of the local remote tracking branch can be different, or it doesn't exist at all. While the race condition does exist, the chance it hits that condition is narrow. vs. this local remote tracking branch not existing can happen permanently (namely GitHub Actions' checkout doesn't have it). So, this is intentionally FETCH_HEAD.

internal/actions/sync_branch.go
```diff
@@ -61,6 +61,10 @@ func SyncBranch(
 		}
 	} else {
 		if opts.Fetch {
+			fetchHead, err := fetchUpstreamCommits(repo, tx, branch)
```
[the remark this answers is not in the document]
draftcode: No. This is actually used outside of that case.

Without this upstream fetch, the merge commit won't exist locally. This makes the rebase process in syncBranchRebase unable to rebase on top of the merged commit.

internal/actions/sync_branch.go
draftcode: Done.

internal/actions/sync_branch.go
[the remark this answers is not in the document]
draftcode: Done.

internal/actions/sync_branch.go
draftcode: Done.

internal/git/log.go
draftcode: Done.

internal/git/log.go
draftcode: Done.

internal/git/log.go
```diff
@@ -0,0 +1,93 @@
...
+
+func trimNUL(s string) string {
+	return strings.Trim(s, "\x00")
+}
+
+// FindClosedPRs finds the "closes #123" instructions from the commit messages. This returns a PR
+// number to commit hash mapping.
+//
+// See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword
+func FindClosedPRs(cis []*CommitInfo) map[int64]string {
```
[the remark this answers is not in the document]
draftcode: Done.

travigd (approved):

PR #221, opened 2024-01-04
[MER-3360] Run av stack tidy on av stack sync

draftcode:
I think this should happen automatically on av stack sync. Otherwise, av
stack sync will result in an error that it cannot find a branch in the
repository.

### DIFF
--- a/cmd/av/stack_sync.go
+++ b/cmd/av/stack_sync.go
@@ -59,6 +59,10 @@ base branch.
 		if err != nil {
 			return err
 		}
+		if _, err = actions.TidyDB(repo, db); err != nil {
+			return err
+		}
+
 		tx := db.WriteTx()
 		defer tx.Abort()
 
--- a/cmd/av/stack_tidy.go
+++ b/cmd/av/stack_tidy.go
@@ -5,8 +5,7 @@ import (
 	"os"
 	"strings"
 
-	"github.com/aviator-co/av/internal/git"
-	"github.com/aviator-co/av/internal/meta"
+	"github.com/aviator-co/av/internal/actions"
 	"github.com/aviator-co/av/internal/utils/colors"
 	"github.com/aviator-co/av/internal/utils/textutils"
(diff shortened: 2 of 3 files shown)


cmd/av/tidy.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"github.com/aviator-co/av/internal/git"
+	"github.com/aviator-co/av/internal/meta"
+)
+
+// tidyDB removes deleted branches from the metadata and returns number of branches removed from the
+// DB.
+func tidyDB(repo *git.Repo, db meta.DB) (int, error) {
+	tx := db.WriteTx()
```
travigd, line 11: Not major, but all of these files (I think) map to actual CLI subcommands (`stack_sync.go` is,,, well,,, `av stack sync`). You could put this in the `actions` package which is our functional catch-all. 🤷

Not something that I think is blocking though.

cmd/av/stack_sync.go
```diff
@@ -59,6 +59,10 @@ base branch.
 		if err != nil {
 			return err
 		}
+		if _, err = tidyDB(repo, db); err != nil {
```
travigd, line 62: I think we should print the `Tidied <n> branches` here (though no need to print that we didn't tidy branches,,, so,,, idk) and also add a flag to disable the tidy behavior (`--no-tidy`). 

This is also not blocking.

travigd (approved):

cmd/av/tidy.go
draftcode: Moved.

cmd/av/stack_sync.go
draftcode: Since tidying is for internal DB to sync up with the git-branches, I feel like it's really internal thing that users shouldn't care about. IDK. Maybe our users do know about av-cli's internal DB and they know what tidy means. In that case, we can surface this and add an option.

PR #254, opened 2024-04-04
#240 - Fix the case the existed PR's Body and PRMetadataCommentEnd can't be compared

Konboi:
To fix https://github.com/aviator-co/av/issues/240


internal/actions/pr.go
```diff
@@ -756,6 +757,12 @@ func readLineUntil(b *bytes.Buffer, line string) error {
 	for {
 		l, err := b.ReadString('\n')
 		if err != nil {
+			// If ReadString encounters an error before finding a delimiter, it returns the data read before the error and the error itself (often io.EOF).
+			// https://pkg.go.dev/bytes#Buffer.ReadString
+			if err == io.EOF && l+"\n" == line {
```
Konboi, line 762: If a PR already exists, the trailing newline gets removed here https://github.com/aviator-co/av/blob/master/internal/actions/pr.go#L344 and it no longer matches with `PRMetadataCommentEnd` https://github.com/aviator-co/av/blob/master/internal/actions/pr.go#L675. What would be a good way to fix this?

Konboi: @draftcode Could you give me advice when you have time?

Konboi: Sorry, I tried to adopt another approach. Please ignore my comment

Issue #243, opened 2024-03-28
Add av stack orphan


draftcode:
Sometimes a user might want to make av/MergeQueue treat a PR as non-stacked. We can add an av stack orphan command to do that (remove the av metadata in the PR description and update the base branch).

Konboi:
I have a question. 

> update the base branch

Does it mean rebase by the parent branch of the branch?

PR #258, opened 2024-04-10
Implement av stack orphan command

Konboi:
Issue: https://github.com/aviator-co/av/issues/243

I tried to implement `av stack orphan` command.

### DIFF
--- a/cmd/av/stack_orphan.go
+++ b/cmd/av/stack_orphan.go
@@ -1,6 +1,10 @@
 package main
 
 import (
+	"fmt"
+	"os"
+	"strings"
+
 	"emperror.dev/errors"
 	"github.com/aviator-co/av/internal/meta"
 	"github.com/spf13/cobra"
@@ -8,7 +12,7 @@ import (
 
 var stackOrphanCmd = &cobra.Command{
 	Use:   "orphan",
-	Short: "Create a pull request not treated av/MergeQueue",
+	Short: "Current branch and the child branches will be orphaned",
 	Args:  cobra.NoArgs,
 	RunE: func(cmd *cobra.Command, _ []string) error {
 		repo, err := getRepo()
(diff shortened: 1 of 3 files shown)


cmd/av/stack_orphan.go
```diff
@@ -0,0 +1,89 @@
...
+			Upstream: upstream,
+			Branch:   branchName,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to run git rebase")
+		}
+
+		if output.ExitCode != 0 {
+			_, _ = fmt.Fprint(os.Stderr,
+				colors.Failure("ERROR: "),
```
Konboi, line 66: Referred https://github.com/aviator-co/av/blob/61844c50acbc15dc0cf153e8b18d16aa196c1a6d/internal/actions/reparent.go#L197-L205

cmd/av/stack_orphan.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to get hash of parent branch: %s", currentBranch.Parent.Name)
+		}
+
+		upstream := currentBranch.Parent.Name
+		if currentBranch.Parent.Trunk {
+			upstream = "remotes/origin/" + currentBranch.Parent.Name
+		}
+
+		output, err := repo.Rebase(git.RebaseOpts{
```
Konboi, line 55: Honestly, I'm not confident that this is the correct way to do a rebase...

draftcode:
Hm. I think I didn't think through the behavior.

What this command should do would be:

1. Find the branch to be orphaned (and the child branches of that branch).
2. Delete the branch metadata from the db (See https://github.com/aviator-co/av/blob/master/internal/actions/tidy.go#L34-L41).

In other words, this command should just delete the av-cli's branch metadata.

Konboi:
Thank you for your reviewing. I'll fix it

Konboi:
I fixed. Is this behaving as you intended? https://github.com/aviator-co/av/pull/258#issuecomment-2064698440

Issue #244, opened 2024-03-28
Add ability to create branch from given branch


draftcode:
tl;dr: add `av branch "<new branch name>" "<base branch name>"` syntax

PR #265, opened 2024-04-26
Support `av stack branch  <new branch> <base branch>`

Konboi:
address #244

```sh
$ av stack branch new-branch parent-branch
```

### DIFF
--- a/cmd/av/stack_branch.go
+++ b/cmd/av/stack_branch.go
@@ -38,7 +38,7 @@ given as the first argument to the command. Branches should only be renamed
 with this command (not with git branch -m ...) because av needs to update
 internal tracking metadata that defines the order of branches within a stack.`,
 	SilenceUsage: true,
-	Args:         cobra.ExactArgs(1),
+	Args:         cobra.RangeArgs(1, 2),
 	RunE: func(cmd *cobra.Command, args []string) (reterr error) {
 		repo, err := getRepo()
 		if err != nil {
@@ -51,6 +51,10 @@ internal tracking metadata that defines the order of branches within a stack.`,
 		}
 
 		branchName := args[0]
+		if len(args) == 2 {
+			stackBranchFlags.Parent = args[1]
+		}
+
 		if stackBranchFlags.Rename {
 			return stackBranchMove(repo, db, branchName, stackBranchFlags.Force)
 		}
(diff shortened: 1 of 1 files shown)


cmd/av/branch.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+
+		tx := db.WriteTx()
+		defer tx.Abort()
+
+		if _, err := repo.CheckoutBranch(&git.CheckoutBranch{Name: baseBranch, NewBranch: false}); err != nil {
+			return err
+		}
+
+		if _, err := repo.CheckoutBranch(&git.CheckoutBranch{Name: newBranch, NewBranch: true}); err != nil {
```
Konboi, line 51: Do we need to back to the working branch? 
With the current implementation, it will checkout a new branch

cmd/av/branch.go
```diff
@@ -0,0 +1,97 @@
...
+			baseHead, err = repo.RevParse(&git.RevParse{Rev: baseBranch})
+			if err != nil {
+				return errors.WrapIff(
+					err,
+					"failed to determine head commit of branch %q",
+					baseBranch,
+				)
+			}
+		}
+		tx.SetBranch(meta.Branch{
```
Konboi, line 67: Can we add a new branch to DB?

e2e_tests/branch_test.go
```diff
@@ -0,0 +1,51 @@
+package e2e_tests
+
+import (
+	"testing"
+
+	"github.com/aviator-co/av/internal/git/gittest"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBranch(t *testing.T) {
```
Konboi, line 10: I'm not sure using the `av stack tree` command is a good way to test. What do you think? Are there any other good ways??

cmd/av/branch.go
```diff
@@ -0,0 +1,97 @@
...
+	"fmt"
+
+	"emperror.dev/errors"
+	"github.com/aviator-co/av/internal/git"
+	"github.com/aviator-co/av/internal/meta"
+	"github.com/spf13/cobra"
+)
+
+var branchCmd = &cobra.Command{
+	Use:     "branch",
```
draftcode, line 13: Sorry. Actually, this should be a part of `av stack branch`.

draftcode (changes requested):

cmd/av/branch.go
Konboi: OK, I'll fix it

Konboi:
Fixed

draftcode (approved):

PR #264, opened 2024-04-26
[WIP] Output stack graph to PR bodies

oleg-codaio:
When creating/updating/syncing stacks, also include the latest stack of changes in each pull request description. This essentially replicates the `av stack tree` visualization in PRs:

image

This is off by default and can be enabled via `pullRequest.writeStack` (for those not yet using a merge queue).

### Work in progress

- [X] Update `av stack sync` to include the current stack in all open PR descriptions
- [X] Update `av pr create` to include the current stack
- [x] Fix bug where `av pr create` doesn't include itself until re-sync
- [x] Fix bug where `av stack submit` doesn't include the stack
- [ ] Figure out how to keep merged PRs included in descendant stacks (until the whole stack tree is merged into trunk)
- [x] Enable setting branch prefix via config

PR #268, opened 2024-05-02
Update av stack submit to open PRs in browser after entire stack is submitted

oleg-codaio:
When run with a stack of unpushed branches, `av stack submit` ends up opening a bunch of new browser windows for each PR one by one, which is a bit disruptive. ~Since we're pushing a single stack, only open the last-created PR once completed.~ Open all the PRs at the end once created instead.

### DIFF
--- a/cmd/av/stack_submit.go
+++ b/cmd/av/stack_submit.go
@@ -69,7 +69,7 @@ If the --current flag is given, this command will create pull requests up to the
 		}
 
 		// ensure pull requests for each branch in the stack
-		var lastCreatedPullRequest *meta.PullRequest
+		createdPullRequestPermalinks := []string{}
 		ctx := context.Background()
 		client, err := getGitHubClient()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/av/stack_submit.go
```diff
@@ -71,13 +72,17 @@ If the --current flag is given, this command will create pull requests up to the
...
+					BranchName:    branchName,
+					Draft:         config.Av.PullRequest.Draft,
+					NoOpenBrowser: true,
 				},
 			)
 			if err != nil {
 				return err
 			}
+			if result.Created {
+				lastCreatedPullRequest = result.Branch.PullRequest
```
oleg-codaio, line 84: An alternative behavior would be to wait until all the PRs are created first, and then bulk-open all new windows at once.

With stack visualizations in https://github.com/aviator-co/av/pull/264, IMO that's unnecessary since the last created PR will have links to all the other stacks.

However, if it's off perhaps it would still make sense to open a bunch of new windows, or alternatively just open the repo to the PR list, e.g., https://github.com/aviator-co/av/pulls.

Thoughts?

oleg-codaio: @draftcode could you let me know what you think?

draftcode: I feel like if we open PR pages (which is default), then it's better to open all (i.e. changing this to bulk open). I usually do not use `av stack submit` (usually use `av pr create), but I want to check the newly created PRs for the last check.

draftcode: Or make this configurable.

oleg-codaio: Alright, for now I updated to just bulk open them all at once at the end. That feels less disruptive than opening one by one and requiring you to keep switching between browser and terminal.

draftcode (approved):

PR #270, opened 2024-05-02
Refactor PR body parsing

oleg-codaio:
As part of adding support for including the stack in PR bodies (see #264), we'll need to extend templating support to handle more than one section:

```
<!-- av pr stack begin --> { stack here } <!-- av pr stack end -->

Here is the main PR body description.

<!-- av pr metadata
This information is embedded by the av CLI when creating PRs to track the status of stacks when using Aviator. Please do not delete or edit this section of the PR.
` ` `
{"parent":"some-parent","parentHead":"abcdef","parentPull":123,"trunk":"main"}
` ` `
-->
```

The current `ParsePRMetadata()` implementation is efficient, but makes it hard to support another section.

This PR updates `ParsePRMetadata` and `AddPRMetadata` to instead make use of substrings and `strings.Builder`. Performance isn't really a concern here - these PR bodies aren't that big and we spend a lot more time interacting with the `git` CLI anyway.

### DIFF
--- a/internal/actions/pr.go
+++ b/internal/actions/pr.go
@@ -1,7 +1,6 @@
 package actions
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


internal/actions/pr.go
```diff
@@ -671,97 +670,78 @@ type PRMetadata struct {
 	Trunk      string `json:"trunk"`
 }
 
-const PRMetadataCommentStart = "<!-- av pr metadata\n"
```
oleg-codaio, line 674: I removed the newline here because I noticed when I'd edit the PR manually on GitHub, I'd end up with a carriage return, i.e., `"<!-- av pr metadata\r\n"` and this parsing would break. Instead, opting to just parse this directly and add any extra whitespace padding explicitly.

draftcode (approved):

PR #271, opened 2024-05-03
Enable writing stack trees to PR bodies

oleg-codaio:
This PR adds a new configuration option, `pullRequest.writeStack`, that if enabled will add stack trees directly to GitHub pull requests.

For ease of navigation, these are added to the top as follows:

image

And when expanded, show the stack of requests from the latest PR down to the trunk branch:

image

Most PR stacks tend to be linear, but an alternate format is used if there are forks in the stack that make it look more like a tree:

image

This isn't the default format since I personally think the one above is cleaner to read, especially on phones that run out of horizontal real estate.

### DIFF
--- a/internal/actions/pr.go
+++ b/internal/actions/pr.go
@@ -905,6 +905,9 @@ func UpdatePullRequestWithStack(
 	}
 
 	body, prMeta, err := ParsePRBody(existingPR.Body)
+	if err != nil {
+		return err
+	}
 
 	newBody := AddPRMetadataAndStack(body, prMeta, branchName, stackToWrite)
 	_, err = client.UpdatePullRequest(ctx, githubv4.UpdatePullRequestInput{
--- a/internal/actions/pr.go
+++ b/internal/actions/pr.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"text/template"
 
@@ -20,6 +21,7 @@ import (
 	"github.com/aviator-co/av/internal/utils/browser"
 	"github.com/aviator-co/av/internal/utils/colors"
 	"github.com/aviator-co/av/internal/utils/sanitize"
+	"github.com/aviator-co/av/internal/utils/stackutils"
 	"github.com/aviator-co/av/internal/utils/stringutils"
 	"github.com/aviator-co/av/internal/utils/templateutils"
 	"github.com/fatih/color"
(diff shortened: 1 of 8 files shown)


internal/actions/pr.go
```diff
@@ -716,15 +724,135 @@ func ReadPRMetadata(body string) (PRMetadata, error) {
...
+		// 1. It actually looks nicer on GitHub
+		// 2. For the Slack GitHub integration, Slack doesn't support and strips out <table> elements in unfurls - we can avoid showing the stack in the unfurl.
+		sb.WriteString("\n<table><tr><td>")
+		sb.WriteString("<details><summary>")
+		if parentPullRequestNumber > 0 {
+			sb.WriteString("<b>Depends on #")
+			sb.WriteString(strconv.FormatInt(parentPullRequestNumber, 10))
+			sb.WriteString(".</b> ")
+		}
+		sb.WriteString("This PR is part of a stack created with <a href=\"https://github.com/aviator-co/av\">Aviator</a>.")
```
oleg-codaio, line 846: Felt like I should include a little advertisement for Aviator in here :) Happy to update as needed though.

internal/config/config.go
```diff
@@ -37,6 +37,10 @@ type PullRequest struct {
 
 	// Branch prefix to use for creating new branches.
 	BranchNamePrefix string
+
+	// If true, the CLI will automatically add/update a comment to all PRs linking other PRs in the stack.
+	// False by default, since Aviator's MergeQueue also adds a similar comment.
+	WriteStack bool
```
oleg-codaio, line 43: I could see this being on by default if there was a way for `av` to auto-detect if `MergeQueue` is being used? Or alternatively if this is turned on then maybe it's not needed in the MergeQueue comments? (I haven't used myself yet.)

internal/actions/pr.go
draftcode: Thanks!

internal/actions/pr.go
```diff
@@ -716,15 +726,135 @@ func ReadPRMetadata(body string) (PRMetadata, error) {
...
 	if err != nil {
 		// No existing metadata comment, so add one.
 		logrus.WithError(err).Debug("could not parse PR metadata (assuming it doesn't exist)")
-		body += "\n\n"
 	}
 
 	sb := strings.Builder{}
+
+	// Don't write out a stack unless there is more than one PR in it.
+	has_multilevel_stack := stack != nil && len(stack.Children) > 0 && len(stack.Children[0].Children) > 0
```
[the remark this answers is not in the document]
oleg-codaio: Oops, done.

draftcode (approved):

draftcode:
I'll cut a new release for av-cli.

Issue #246, opened 2024-03-28
av stack sync --prune Left in detached head on branch


draftcode:
Instead we should checkout trunk, rather than leaving user in a detached head.

PR #273, opened 2024-05-08
Checkout a trunk instead of switching to the detached HEAD

Konboi:
Addressed https://github.com/aviator-co/av/issues/246


internal/actions/sync_stack.go
```diff
@@ -190,8 +190,9 @@ func SyncStack(ctx context.Context,
 				"  - deleting merged branch ", colors.UserInput(currentBranch),
 				"\n",
 			)
-			if _, err := repo.Git("switch", "--detach"); err != nil {
-				return errors.Errorf("cannot switch to detached HEAD: %v", err)
+
+			if _, err := repo.CheckoutBranch(&git.CheckoutBranch{Name: br.Parent.Name}); err != nil {
+				return errors.Errorf("cannot checkout to parent branch %s: %v", br.Parent.Name, err)
```
Konboi, line 195: You mentioned 

> Instead we should checkout trunk

in https://github.com/aviator-co/av/issues/246

Does "trunk' refer to something different from a 'parent branch' or same?

draftcode: Trunk is typically 'main' or 'master'. It'll be the root of those stacked branches.

Konboi: I see, I'll fix it

draftcode (approved):

PR #296, opened 2024-05-20
Handle non-terminal output

draftcode:

### DIFF
--- a/cmd/av/stack_switch.go
+++ b/cmd/av/stack_switch.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"os"
 	"strings"
 
+	"emperror.dev/errors"
 	"github.com/aviator-co/av/internal/actions"
 	"github.com/aviator-co/av/internal/git"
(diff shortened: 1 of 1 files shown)


cmd/av/stack_switch.go
```diff
@@ -44,6 +47,10 @@ var stackSwitchCmd = &cobra.Command{
 		for _, node := range rootNodes {
 			branchList = append(branchList, stackSwitchBranchList(repo, tx, branches, node)...)
 		}
+
+		if !isatty.IsTerminal(os.Stdout.Fd()) {
```
jainankit, line 51: What's the scenario here, where would this not work?

jainankit (approved):

cmd/av/stack_switch.go
draftcode: For example, GitHub Actions runner will fail because it doesn't have a terminal attached / cannot allocate any tty.

PR #298, opened 2024-05-21
Run mock GitHub server for testing

draftcode:
This allows us to test stack sync with the default option.


internal/gh/pullrequest.go
```diff
@@ -9,17 +9,11 @@ import (
 )
 
 type PullRequest struct {
-	ID     string
```
draftcode, line 12: Note that, this is just deleting unused fields.

jainankit (approved):

PR #318, opened 2024-06-07
Prevent `av commit amend` and `av commit create` for merged branches …

Brookke:
…#312

## Todo

- [x] add tests
- [x] sync branch state before checking


cmd/av/commit_amend.go
```diff
@@ -57,6 +60,13 @@ func runAmend(repo *git.Repo) error {
 		commitArgs = append(commitArgs, "--message", commitAmendFlags.Message)
 	}
 
+	tx := db.ReadTx()
+	branch, _ := tx.Branch(currentBranch)
+	// Todo: do we need to refresh the state of the branch
+	if branch.PullRequest != nil && branch.PullRequest.State == githubv4.PullRequestStateMerged {
```
Brookke, line 66: @draftcode am I correct in assuming we need to run `actions.SyncBranch` before checking this?

Brookke: actually just found `UpdatePullRequestState`

Brookke:
Unsure how best to test this given it uses the github api. I have at least made sure that the commands pass the tests by not blocking the command if UpdatePullRequestState fails, this also ensures the CLI can be used offline

cmd/av/commit_create.go
```diff
@@ -57,6 +61,24 @@ func runCreate(repo *git.Repo) error {
...
+
+	client, err := getGitHubClient()
+	if err != nil {
+		return err
+	}
+
+	ctx := context.Background()
+	prUpdateResult, err := actions.UpdatePullRequestState(ctx, client, writeTx, currentBranch)
+	if err != nil {
+		fmt.Fprint(os.Stderr, colors.Warning("failed to check pull request state, continuing with commit"), "\n")
```
Brookke, line 75: this doesn't get outputted

cmd/av/commit_amend.go
```diff
@@ -57,6 +61,25 @@ func runAmend(repo *git.Repo) error {
...
+	writeTx := db.WriteTx()
+	defer writeTx.Abort()
+
+	client, err := getGitHubClient()
+	if err != nil {
+		return err
+	}
+
+	ctx := context.Background()
+	prUpdateResult, err := actions.UpdatePullRequestState(ctx, client, writeTx, currentBranch)
```
Brookke, line 73: should this fallback to the offline state if this fails

cmd/av/commit_amend.go
```diff
@@ -57,6 +61,26 @@ func runAmend(repo *git.Repo) error {
 		commitArgs = append(commitArgs, "--message", commitAmendFlags.Message)
 	}
 
+	tx := db.WriteTx()
+	defer tx.Abort()
+
+	client, err := getGitHubClient()
```
draftcode, line 67: I think that making GitHub API call in amend is too much. Can you skip this part and just rely on the local information (e.g. MergeCommit existence)?

Brookke: Whats the concern with calling the API? Too slow? Happy to switch to the local info, but was thinking chances are the local information wont always be up to date so it will not really fix the problem.

draftcode: Slowness and expectation. I think that people would not expect `git commit` to do any network interaction. Keeping the same expectation level for these command would make less surprise. I understand that this wouldn't completely prevent the accidental amend, but that's fine and expected.

Brookke: Have removed this in https://github.com/aviator-co/av/pull/318/commits/d2d97d9630a7200950c7856f130aa9caf95c0522 to unblock this pr, however while I understand the philosophy I think its a worse experience - I wonder if instead it should be configurable. Is there a command that allows you to update the status of all pr's without having to do a full sync of the stacks?

draftcode (approved):

Issue #9, opened 2022-04-22
Add support for remotes other than origin


twavv:
Most code assumes the remote name we care about is `origin`. This is usually true, but might not always be.

PR #340, opened 2024-06-15
Support for remotes other than `origin`

Konboi:
https://github.com/aviator-co/av/issues/9

Now `av` supports only remote `origin` but I sometimes set multiple remotes and set a name something other than `origin`

To support it, I added `remote` and updated to use it


internal/git/git.go
```diff
@@ -108,6 +110,14 @@ func (r *Repo) TrunkBranches() ([]string, error) {
 	return branches, nil
 }
 
+func (r *Repo) GetRemoteName() string {
+	if config.Av.Remote != "" {
```
Konboi, line 114: Which is better, setting it as the initial value (at [here](https://github.com/aviator-co/av/pull/340/files#diff-54c7c1af5fa8d5db4dc49f0e8e80e93ba2b1183ba4d5c9e2e5729e6deae6a3cdR71)) or checking it here?

draftcode: This is fine.

draftcode (approved):

PR #344, opened 2024-06-17
Improve the stack branch command to handle trunk branches as parents

draftcode:
When creating a new branch from a trunk branch, the stack branch command
is better to start off from the remote tracking branch instead of the
local branch. This way, the new branch will be created from the latest
state of the trunk branch.

While we are here, accept `origin/HEAD` as the parent branch name and
resolve it to the default branch.

Previously, this command automatically "adopts" the parent branch to av
automatically, assuming that the parent branch's parent is the default
branch, which might not be true. Instead, this command checks if the
parent branch is already adopted or not.

### DIFF
--- a/cmd/av/main.go
+++ b/cmd/av/main.go
@@ -172,8 +172,6 @@ func discoverGitHubAPIToken() string {
 	return ""
 }
 
-var errNoGitHubToken = errors.Sentinel("No GitHub token is set (do you need to configure one?).")
-
 func getGitHubClient() (*gh.Client, error) {
 	token := discoverGitHubAPIToken()
 	if token == "" {
--- a/cmd/av/render_error.go
+++ b/cmd/av/render_error.go
@@ -1,11 +1,16 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 
+	"emperror.dev/errors"
 	"github.com/charmbracelet/glamour"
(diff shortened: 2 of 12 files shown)


jainankit:
A few questions (user experience related):
- With this new functionality, would we internally do a git fetch for that remote ref in case our local trunk branch is not up to date?
- A separate minor UX note that now the behavior is slightly different for trunk branch vs feature branches. We should show this as a message to the user when running av stack branch on trunk.
- If we report an error when parent is not adopted, we should recommend user with the command to adopt the entire stack. There could be scenario where the user wants to push multiple branches together. Or there could be a param within av stack branch to support that behavior.

draftcode:
> A few questions (user experience related):
> 
> * With this new functionality, would we internally do a git fetch for that remote ref in case our local trunk branch is not up to date?

Answered offline, but this is using the remote tracking branches, which is a local ref. It doesn't do any git-fetch.

> * A separate minor UX note that now the behavior is slightly different for trunk branch vs feature branches. We should show this as a message to the user when running av stack branch on trunk.

This is consistent with av-stack-sync behavior.

> * If we report an error when parent is not adopted, we should recommend user with the command to adopt the entire stack. There could be scenario where the user wants to push multiple branches together. Or there could be a param within av stack branch to support that behavior.

I will update the error message.

jainankit (approved):

PR #351, opened 2024-06-25
Do not push a branch if an ancestor doesn't have a PR

draftcode:
https://linear.app/mergequeue/issue/MER-4307/swap-the-order-of-push-pr-update-to-avoid-the-unintended-ci-triggers


jainankit:
What's the rationale behind this change?

jainankit (approved):

PR #401, opened 2024-08-21
Use the golangci-lint GitHub Action

draftcode:
It seems that there's a change in golangci-lint that makes CI runs fail.

https://github.com/aviator-co/av/pull/399

Use the official GitHub Action for golangci-lint instead of installing
it manually.

https://github.com/aviator-co/av/issues/402


alex-aviator (approved):

internal/gh/pullrequest.go
```diff
@@ -88,7 +88,7 @@ type GetPullRequestsInput struct {
 	HeadRefName string
 	BaseRefName string
 	States      []githubv4.PullRequestState
-	First       int64
+	First       int32
```
jainankit, line 91: why?

jainankit (approved):

internal/gh/pullrequest.go
draftcode: Because the lint says so...

Issue #396, opened 2024-08-14
`av stack sync` won't update the PR metadata if the commit hashes are aligned


draftcode:
https://github.com/aviator-co/av/blob/81f6fc4142623c8ac2b71ffbba6ebf549904bd0c/internal/gh/ghui/push.go#L449-L455

In addition to hashes, it should check if the PR metadata matches with the local branch metadata.

PR #403, opened 2024-08-22
Check local metadata is updated compared with PR's metadata

Konboi:
addressed https://github.com/aviator-co/av/issues/396

### DIFF
--- a/internal/gh/ghui/push.go
+++ b/internal/gh/ghui/push.go
@@ -31,13 +31,12 @@ const (
 	continuePush = "Yes. Push the branches to GitHub."
 	abortPush    = "No. Do not push the branches to GitHub."
 
-	reasonAlreadyUpToDate    = "Already up-to-date."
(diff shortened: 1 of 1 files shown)


internal/gh/ghui/push.go
```diff
@@ -453,6 +443,15 @@ func (vm *GitHubPushModel) calculateChangedBranches() tea.Msg {
 			})
 			continue
 		}
+
+		if !isDifferencePRMetadata(avbr, vm) {
```
Konboi, line 447: > In addition to hashes, it should check if the PR metadata matches with the local branch metadata. 
https://github.com/aviator-co/av/issues/396

Does this mean that checking the metadata with the hash, together? Or does it check only this case?? 
I implemented it as the latter. But if I'm wrong please let me know, I'll fix it

draftcode: I meant first.

Issue #388, opened 2024-08-07
Check `av stack sync` or `av stack restack` or `git rebase` is in progress


draftcode:
We should prevent certain commands from running when there's an ongoing sync / rebase / restack / reorder.

PR #410, opened 2024-08-30
Check av's state file and `REBASE_HEAD` to prevent conflict

Konboi:
Addressed #388

### DIFF
--- a/internal/git/rebase.go
+++ b/internal/git/rebase.go
@@ -1,11 +1,17 @@
 package git
 
 import (
+	"fmt"
+	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
 
 	"github.com/sirupsen/logrus"
 )
+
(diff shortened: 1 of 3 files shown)


internal/git/rebase.go
```diff
@@ -1,12 +1,18 @@
...
+	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
 
 	"github.com/sirupsen/logrus"
 )
 
+// ref: https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt-codeREBASEHEADcode
+const REBASE_STATE_FILE = "REBASE_HEAD"
```
Konboi, line 14: In other places, `REBASE_HEAD` is defined as string directly. 
Should I align with that here as well? It means that I shouldn't define this string as const?

Issue #395, opened 2024-08-14
Updating a PR metadata before git-push is not possible in some cases


draftcode:
In https://github.com/aviator-co/av/pull/352, we switched to call GitHub API to update the PR metadata first, then doing a git-push. This is to make sure that a CI service sees the base branch data correctly.

However, if a branch order is swapped, the PR metadata cannot be updated. When a stack is like `main <- feature1 <- feature2` originally, and then swapped the feature1 and feature2 (`main <- feature2 <- feature1`), without a git-push beforehand, GitHub recognizes that there's no new commit between feature2 and feature1 (because GH is still recognizing them based on the pre-reorder state), rejecting the API request.

draftcode:
I guess we can check if the branch order is the same as the current local branch order. This won't detect all the cases that GH would recognize there's no new commit, but should cover some.

wahyudibo:
Hi @draftcode, i'm pleased to meet you. I've been assigned to this issue by @jainankit and trying to replicate the issue on my end. I have followed the video in https://docs.aviator.co/aviator-cli and here's the result by running `av stack tree`

```
$ av stack tree

  * frontend-change
  │ https://github.com/wahyudibo/test-aviator-av/pull/2
  │
  * backend-1 (HEAD)
  │ https://github.com/wahyudibo/test-aviator-av/pull/1
  │
  * main
```

I run these commands to swap the branches order

```
$ av stack reorder
Starting branch frontend-change at c8ba99a
  - applied commit 50332e7 without conflict (HEAD is now at 2dd3ce4)
Starting branch backend-1 at 2dd3ce4
  - applied commit c54ea93 without conflict (HEAD is now at 1fbbf80)
  - applied commit 934a4d3 without conflict (HEAD is now at 8d8b41d)
Reorder complete!

The stack was reordered successfully.

$ av stack tree

  * backend-1 (HEAD)
  │ https://github.com/wahyudibo/test-aviator-av/pull/1
  │
  * frontend-change
  │ https://github.com/wahyudibo/test-aviator-av/pull/2
  │
  * main
```

Then i run `av stack sync` and choose `Yes` when it asked me to push to github and i got this result

```
$ av stack sync



  ✓ GitHub fetch is done
  ✓ Restack is done

    * ✓ backend-1 80669e1
    │
    * frontend-change eb81b3e
    │
    * main 312a571

  ⡿ Pushing to GitHub...

    Following branches are being pushed...

      frontend-change
[...]

PR #436, opened 2024-10-23
fix issue with updating PR when the stack is reordered (#395)

wahyudibo:
This PR addresses issue #395

# Fixing Steps

1. Handle the error returned when updating a PR before the local branch is pushed to GitHub by marking the `reRunPRUpdate` in `GitHubPushModel` as true and continue to the next iteration.
2. After `runGitPush()`, check `reRunPRUpdate` value. If true, then re-run `updatePRs` again to fix the PR base branch

# Before 
Previously, when the branches were reordered and we ran `av stack sync` it would return `There are no new commits between base branch 'base' and head branch 'head'`, and the process is stopped

# After
The previous issue did not happen and the PR order is now in sync with the local machine's re-ordered stack.

Image 2024-10-23-11-28-33

### DIFF
--- a/internal/gh/ghui/push.go
+++ b/internal/gh/ghui/push.go
@@ -37,6 +37,8 @@ const (
 	reasonPRIsClosed        = "PR is closed."
 	reasonParentNotPushed   = "Parent branch is not pushed to remote."
 	reasonNoPR              = "Some branches in a stack do not have a PR."
+
+	ghErrorResponseNoNewCommits = "There are no new commits between base branch"
 )
 
 type pushCandidate struct {
@@ -101,6 +103,7 @@ type GitHubPushModel struct {
 	calculatingCandidates bool
 	askingForConfirmation bool
 	runningGitPush        bool
+	reRunPRUpdate         bool
 	done                  bool
 }
 
@@ -274,6 +277,11 @@ func (vm *GitHubPushModel) runUpdate() (ret tea.Msg) {
 	if err := vm.runGitPush(); err != nil {
 		return err
 	}
+	if vm.reRunPRUpdate {
+		if err := vm.updatePRs(ghPRs); err != nil {
+			return err
+		}
+	}
 	return &GitHubPushProgress{gitPushDone: true}
 }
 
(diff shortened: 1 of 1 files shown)


draftcode:
It seems there is a bug.

```
set -ex

git checkout main
av stack branch branch-1
git commit --allow-empty -m "Commit on branch-1"
av pr create --title "PR on branch-1" --body "PR on branch-1"

git checkout main
av stack branch branch-2
git merge --ff-only branch-1
av pr create --title "PR on branch-2" --body "PR on branch-2"

git checkout branch-2
av stack reparent --parent branch-1

av stack sync --all --push yes
```

With this change, the last `av stack sync` exits with 0 saying it's "pushed", but it's actually failing since there's no new commit in branch-2. As a result, `av stack sync` becomes no-op and repeatedly running it will just trying to push, and it says it's "pushed", but it didn't. Prior to this change, it correctly errors out with there's no new commit.

wahyudibo:
Thank you for checking @draftcode , i'll try to reproduce the issue on my end and fix it

wahyudibo:
Hi @draftcode, i tried to reproduce the issue on my end, and here's what I've got:

Image 2024-10-24-16-09-41

my understanding of the issue #395 is, we have this error `There are no new commits between base branch 'base' and head branch 'head'` which prevents us from updating the PR when the stack is re-ordered. So my steps to fix this issue are:

1. catch the above error into a conditional statement (if)
2. inside the if block, remark this session to re-run pr update after git push
3. continue to the next iteration of pr checks

but from the example above, seems you're expecting the error to appear and stop the process. So what is the condition that is required to differentiate between this case and the case from #395? because both have the same `there's no new commit` error but we want the process to be continued in #395 but stopped in this case. Please kindly suggest.

draftcode:
https://github.com/aviator-co/av/pull/436#issuecomment-2432946568 truely do not have any new commit between the parent and its child, so it should be errored out. https://github.com/aviator-co/av/issues/395 is that depending on the order of the operations, it appears as there's no new commit between the parent and its child, but it's just a transient case.

Locally, it should be able to tell if there is no new commit or not. Traversing or doing something like https://github.com/aviator-co/av/issues/395#issuecomment-2289724898.

Issue #440, opened 2024-10-28
Branch metadata update check is not working


draftcode:
With https://github.com/aviator-co/av/pull/403, we started checking the PR metadata alignment.

However, the change uses `vm.getPRs()` which at this point won't return the PR metadata for the given PR, which results in all PRs are considered as "modified". So this is not working as expected.

PR #448, opened 2024-11-05
Issue #440

Konboi:
Before I used `getPRs` method ref: (https://github.com/aviator-co/av/blob/master/internal/gh/ghui/push.go#L323)
However, the method only returns push candidates branches ref: (https://github.com/aviator-co/av/blob/master/internal/gh/ghui/push.go#L325). In my understanding correct, when target branch is a push candidate branch, this logic will work well. But if the target branch is **not** a push candidate branch, it won't work well.

So, I updated to check PR directly.

### DIFF
--- a/internal/gh/ghui/push.go
+++ b/internal/gh/ghui/push.go
@@ -97,6 +97,8 @@ type GitHubPushModel struct {
 	pushCandidates []pushCandidate
 	noPushBranches []noPushBranch
 	pushPrompt     *selection.Model[string]
+	// key is  internal/gh/pullrequest.PullRequest.ID
+	pullRequestsCache map[string]*gh.PullRequest
 
 	calculatingCandidates bool
 	askingForConfirmation bool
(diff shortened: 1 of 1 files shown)


internal/gh/ghui/push.go
```diff
@@ -543,26 +543,15 @@ func createPRMetadata(branch meta.Branch, vm *GitHubPushModel) actions.PRMetadat
 }
 
 // Compare local metadata with PR metadata for any changes
+// If something error occurs, return true to be safe
 func isDifferencePRMetadata(avbr meta.Branch, vm *GitHubPushModel) bool {
 	local := createPRMetadata(avbr, vm)
 
-	prs, err := vm.getPRs()
+	pr, err := vm.client.PullRequest(context.Background(), avbr.PullRequest.ID)
```
draftcode, line 550: The direction looks good. Can you change the logic slightly so that we don't have to make the API call twice (once in this and another in getPRs)?

Konboi: Sorry for the delayed response. I updated

draftcode (approved):

draftcode:
Can you rebase this to master?

draftcode (approved):

Issue #409, opened 2024-08-30
Interactive option for `av stack restack`


dhaus67:
Sometimes, when working with stacked PRs there might be commits which I want to squash / edit before pushing them to remote.

It would be nice if `av stack restack` would allow to also support to interactively rebase the commits on each branch and allow the usual suspects of squashing, editing commits etc.

Alternatively, the missing options could also be added to `av stack reorder` but the `restack` command feels like the more intuitive option to me comparing it to standard git operations.

draftcode:
I think that we can do the same check as --autosquash and prompt the user to do the squash. So it's like:

```
   Commits included in branch `feature-xyz` has an autosquash keyword.
   
   * Implement XYZ
   * amend!: fix a typo
   * amend!: regenerate autogen files
   
   Do you want to autosquash these commits? [y|n]
```

PR #451, opened 2024-11-07
Issue #409 - WIP

Chafid:
This is for handling the issue raised in https://github.com/aviator-co/av/issues/409.
Interactive menu has been added but restack still not working properly yet. Putting this as WIP

### DIFF
--- a/internal/git/rebase.go
+++ b/internal/git/rebase.go
@@ -1,9 +1,14 @@
 package git
 
 import (
+	"bytes"
+	"os"
+	"os/exec"
 	"regexp"
 	"strings"
 
+	"emperror.dev/errors"
+	tea "github.com/charmbracelet/bubbletea"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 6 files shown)


draftcode:
I think that there's a bit of misunderstanding. All you need would be add an interactive option to RebaseOpts at https://github.com/aviator-co/av/blob/ed184fe8384f62dc1ee8f2328ffc16d1350e4473/internal/sequencer/sequencer.go#L202, and run git rebase interactively so that users can get a chance to squash.

Chafid:
Ok, I'm working on this

internal/git/rebase.go
```diff
@@ -49,6 +52,15 @@ func (r *Repo) Rebase(opts RebaseOpts) (*Output, error) {
 			Args: []string{"rebase", "--skip"},
 		})
 	}
+
+	if opts.Interactive {
+		args = append(args, "-i")
+		args = append(args, "master")
```
draftcode, line 58: I don't think you need this. You should just add "-i" and run in interactive. Other args are added in other parts of the code.

Chafid: got it. I've pushed the changes, but I still have issue with the text editor. Seems like the fancy text that show up during non interactive is still showing up during interactive session. I've sent you the image on discord

draftcode: Hm. I think you would need to change this to use https://pkg.go.dev/github.com/charmbracelet/bubbletea#Exec or https://pkg.go.dev/github.com/charmbracelet/bubbletea#ExecProcess. Bubbletea tries to control the tty, so we should stop it while executing another process.

PR #452, opened 2024-11-08
feat: add --all flag to av commit amend

dacuna-ic:
Adds the --flag to `av commit amend` to match `av commit create`

### DIFF
--- a/cmd/av/commit_amend.go
+++ b/cmd/av/commit_amend.go
@@ -96,5 +96,5 @@ func init() {
 		BoolVar(&commitAmendFlags.NoEdit, "no-edit", false, "amend a commit without changing its commit message")
 	commitAmendCmd.MarkFlagsMutuallyExclusive("message", "no-edit")
 	commitAmendCmd.Flags().
-		BoolVarP(&commitCreateFlags.All, "all", "a", false, "automatically stage modified files (same as git commit --all)")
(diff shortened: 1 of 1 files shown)


cmd/av/commit_amend.go
```diff
@@ -56,6 +59,9 @@ func runAmend(repo *git.Repo, db meta.DB) error {
 	if commitAmendFlags.NoEdit {
 		commitArgs = append(commitArgs, "--no-edit")
 	}
+	if commitCreateFlags.All {
```
draftcode, line 62: I think this should be commitAmendFlags

cmd/av/commit_amend.go
```diff
@@ -89,4 +95,6 @@ func init() {
 	commitAmendCmd.Flags().
 		BoolVar(&commitAmendFlags.NoEdit, "no-edit", false, "amend a commit without changing its commit message")
 	commitAmendCmd.MarkFlagsMutuallyExclusive("message", "no-edit")
+	commitCreateCmd.Flags().
```
draftcode, line 98: This should be commitAmendCmd.

draftcode (changes requested):

dacuna-ic:
@draftcode should be good now! sorry for that, just copied stuff in a rush and didn't pay much attention 😅

PR #492, opened 2024-12-13
wip: add squash command

Brookke:
This is a work in progress, opening for input on the open questions I have around output. Maybe this is fine as is...

### DIFF
--- a/cmd/av/squash.go
+++ b/cmd/av/squash.go
@@ -1,0 +1,89 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"emperror.dev/errors"
+	"github.com/aviator-co/av/internal/git"
+	"github.com/aviator-co/av/internal/utils/colors"
+	"github.com/shurcooL/githubv4"
+	"github.com/spf13/cobra"
+)
+
+var squashCmd = &cobra.Command{
+	Use:   "squash",
+	Short: "Squash commits of the current branch into a single commit",
+	Args:  cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		repo, err := getRepo()
+		if err != nil {
+			return err
+		}
+
+		db, err := getDB(repo)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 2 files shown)


cmd/av/squash.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			return err
+		}
+
+		branch, branchExists := db.WriteTx().Branch(currentBranchName)
+		if !branchExists {
+			return errors.New("current branch does not exist in the database")
+		}
+
+		// Could this be abstracted into a helper function?
```
Brookke, line 53: I think this would be useful as a generic repo command. Lmk what you think

cmd/av/squash.go
```diff
@@ -0,0 +1,89 @@
...
+
+		if _, err := repo.Git("reset", "--soft", firstCommitSha); err != nil {
+			return err
+		}
+
+		if _, err := repo.Git("commit", "--amend", "--no-edit"); err != nil {
+			return err
+		}
+
+		// What should the output be?
```
Brookke, line 85: Should there be any output? maybe its fine without

cmd/av/squash.go
```diff
@@ -0,0 +1,89 @@
...
+		// Should a sync be triggered?
```
Brookke, line 86: should I be running something like runPostCommitRestack here?

Brookke:
Hey @draftcode @jainankit, Just realised I never got this merged, are you able to help with the above questions? Then I will address then ready for merge

cmd/av/squash.go
Brookke: Added :)

tulioz:
https://github.com/user-attachments/assets/66644b12-c251-4508-8766-80958cb57c0e

When running locally the squash happens correctly, but process also hangs

tulioz (approved):
lgtm! sorry for not including `--debug` output, screen recording was probably not the most useful. Congrats on the new role!

PR #495, opened 2024-12-21
command: av cherry-pick dev work

### LINKED ISSUE #248: Add av cherry-pick

kolharsam:
This PR will resolve #248 

This is still very much a WIP but opening the initial work for feedback and also to help set expectations on what we need to do for this.

### DIFF
--- a/cmd/av/cherry_pick.go
+++ b/cmd/av/cherry_pick.go
@@ -1,0 +1,200 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/aviator-co/av/internal/git"
+	"github.com/aviator-co/av/internal/meta"
+	"github.com/aviator-co/av/internal/utils/cleanup"
+	"github.com/aviator-co/av/internal/utils/colors"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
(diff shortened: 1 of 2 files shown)


tulioz (commented):
Hey @kolharsam thanks for starting on this issue! My one big comment is that we tend to be a lot more opinionated on what flags we offer users versus the base `git` commands. From a quick skim, I think `--edit` and `--no-commit` are the only two we may want to support outside of abort/continue/skip/quit.

kolharsam:
> Hey @kolharsam thanks for starting on this issue! My one big comment is that we tend to be a lot more opinionated on what flags we offer users versus the base `git` commands. From a quick skim, I think `--edit` and `--no-commit` are the only two we may want to support outside of abort/continue/skip/quit.

Hey @tulioz thanks for the response! 
I'll take note of this as I build it all out completely...but I also wanted to point you to this comment by @jainankit 
https://github.com/aviator-co/av/issues/248#issuecomment-2408244060

Let me know what you think! Thanks!

tulioz:
Sorry for the delayed response, we were on winter break this past week. I definitely agree that we don't want to be blocking users, and that the absolute must have flags are: `--continue | --skip | --abort | --quit`, since those ensure users can get out of bad states. From the optional parameters, most of those flags are extreme edge cases outside of `--edit` and `--no-commit`, but even those are lower priority than `--continue | --skip | --abort | --quit`.

PR #556, opened 2025-04-16
Make pre-av hooks interactive and exit on error.

amartani:
Interactive for better user experience (show progress of the hook as it is running) and exit on error so that checks that fail actually stop the command.

### DIFF
--- a/cmd/av/sync.go
+++ b/cmd/av/sync.go
@@ -337,7 +337,9 @@ func (vm *syncViewModel) initSync() tea.Cmd {
 	}
 	return func() tea.Msg {
 		output, err := vm.repo.Run(&git.RunOpts{
-			Args: []string{"hook", "run", "--ignore-missing", "pre-av-sync"},
+			Args:        []string{"hook", "run", "--ignore-missing", "pre-av-sync"},
+			Interactive: true,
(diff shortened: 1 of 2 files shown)


cmd/av/sync.go
```diff
@@ -337,7 +337,9 @@ func (vm *syncViewModel) initSync() tea.Cmd {
 	}
 	return func() tea.Msg {
 		output, err := vm.repo.Run(&git.RunOpts{
-			Args: []string{"hook", "run", "--ignore-missing", "pre-av-sync"},
+			Args:        []string{"hook", "run", "--ignore-missing", "pre-av-sync"},
+			Interactive: true,
+			ExitError:   true,
```
draftcode, line 342: Unlike `av pr` equivalent, this "Interactive" option might not work as expected. Because the terminal is under bubbletea's control in this code, the program exec-ed from this part wouldn't be able to control the terminal. For this reason, bubbletea has a special construct to pause its terminal control and exec the specified program https://pkg.go.dev/github.com/charmbracelet/bubbletea#ExecProcess

It's fine for now, but just let you know. I can follow up later.

draftcode (approved):

PR #574, opened 2025-07-30
make squash command add some visual feedback that the squash is running

tulioz:

### DIFF
--- a/cmd/av/squash.go
+++ b/cmd/av/squash.go
@@ -4,12 +4,16 @@ import (
 	"context"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


tulioz:
@Brookke making some small changes to the squash command, would love your feedback as well if you have any thoughts.

draftcode (approved):

Brookke:
Looks good!


However, realised I have a bug in this and the current prod version where it squashes one too many commits 🤦🏻 I probably missed this in my PR due to my too simple testing setup. @tulioz

PR #579, opened 2025-08-13
fix: prevent squash on out-of-sync branches

Brookke:
## Summary
- Add validation to ensure branches are in sync with their parent before allowing squash operations
- Prevent data loss when parent branches have been force-pushed or modified externally

## Changes
- Added `validateBranchSync` function with dual validation checks (parent head movement and merge base verification)
- Updated squash command to validate sync before proceeding
- Added comprehensive test cases covering force-push and out-of-sync scenarios

### DIFF
--- a/cmd/av/helpers.go
+++ b/cmd/av/helpers.go
@@ -165,3 +165,47 @@ func branchNameArgs(
 	branches, _ := allBranches(cmd.Context())
 	return branches, cobra.ShellCompDirectiveNoSpace
 }
+
+// validateBranchSync checks if the current branch is in sync with its parent branch.
+// Only enforces this check for non-trunk branches when we have a stored parent head.
+// This validation prevents commands from operating on branches that are out of sync
+// with their parent, which could lead to unexpected behavior or loss of commits.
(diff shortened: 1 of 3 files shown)


cmd/av/helpers.go
```diff
@@ -165,3 +165,47 @@ func branchNameArgs(
 	branches, _ := allBranches(cmd.Context())
 	return branches, cobra.ShellCompDirectiveNoSpace
 }
+
+// validateBranchSync checks if the current branch is in sync with its parent branch.
+// Only enforces this check for non-trunk branches when we have a stored parent head.
+// This validation prevents commands from operating on branches that are out of sync
+// with their parent, which could lead to unexpected behavior or loss of commits.
+func validateBranchSync(ctx context.Context, repo *git.Repo, branch meta.Branch) error {
```
Brookke, line 173: Not sure the philosophy around how much this CLI should act as a wizard. I.e. rather than throwing an error maybe the cli should prompt for "do you want to run av sync?". 

Happy to update depending on what you prefer

PR #641, opened 2026-01-07
Fix: Prevent incorrect merge propagation when PR is flattened

### LINKED ISSUE #640: Merging a downstream PR (without parents) causes av to think the entire stack is merged (and corrupts its state)
Easy example, consider a stack with three PRs in it (main <- PR 1  <- PR 2 <-  PR 3)

Say you take the changes from PR 3 and flatten them into PR 2 (or whatever upstream PR):
- GitHub will mark the PR as merged (because it transitioned to no diff)
- running `av sync` erroneously thinks _all_ PRs in the stack are now merged (puts `mergeCommit` into their state, though `pullRequest.state` is still `OPEN`), and tries to clean them up and truncates the rest of the stack

---

Workaround: ctrl-c ASAP to stop it from cleaning your branches/PRs up, and then:
- remove the `mergeCommit` state from `.git/av/av.db` for all the branches that shouldn't be flagged merged.
- remove the merged PR completely from state; or this will reoccur

btjfsu:
## Summary
Fixes #640

When a downstream PR is flattened into its parent (causing GitHub to mark it as "merged" with no diff), `av sync` was incorrectly propagating the merge status to all parent PRs in the stack, corrupting state.

## Changes
- Verify merge commits are in trunk history before propagating upward
- Add merge-base ancestry check in `updateMergeCommitsFromChildren()`
- Only propagate merge status when commit is reachable from remote trunk
- Add explicit MERGED state check in `UpdatePullRequestState` for clarity

## How it Works
The fix adds a verification step before propagating merge commits from child branches to parents. It uses `git merge-base --is-ancestor` to check if the merge commit is actually reachable from the remote trunk. This prevents the scenario where flattening PR 3 into PR 2 causes av to think PR 1 and PR 2 are also merged.

### DIFF
--- a/internal/utils/uiutils/tea.go
+++ b/internal/utils/uiutils/tea.go
@@ -18,7 +18,7 @@ type BubbleTeaModelWithExitHandling interface {
 
 func RunBubbleTea(model BubbleTeaModelWithExitHandling) error {
 	var opts []tea.ProgramOption
-	if !isatty.IsTerminal(os.Stdout.Fd()) {
+	if !isatty.IsTerminal(os.Stdin.Fd()) || !isatty.IsTerminal(os.Stdout.Fd()) {
 		opts = []tea.ProgramOption{
 			tea.WithInput(nil),
 		}


tulioz:
Thanks for opening a PR to fix this! Sorry for not responding last week we were quite busy coming back from break, I'll set some time aside this week to review.

PR #653, opened 2026-01-22
respect branchNamePrefix config in all branch creation paths

willschnicke-silkline:
The branchNamePrefix config setting was only being used when auto-generating
branch names via `av commit -b`, but not when using `av branch` directly or
`av branch --split` to split commits.

Now the prefix is applied consistently across all branch creation commands by
centralizing the logic in a single `applyBranchNamePrefix` helper function.
When branches are created (whether explicitly named or auto-generated), the
configured prefix gets prepended.

## Changes
- Added `applyBranchNamePrefix()` helper function to centralize prefix logic
- Applied prefix in `createBranch()` for `av branch <name>`
- Applied prefix in `branchSplit()` for `av branch --split`
- Removed duplicate prefix logic from `branchNameFromMessage()` in commit.go
- Updated test repos to explicitly set empty branchNamePrefix to avoid interference from global config

## Testing
- All existing tests pass
- Added comprehensive test coverage for branch name prefix behavior across all creation paths

### DIFF
--- a/internal/actions/pr.go
+++ b/internal/actions/pr.go
@@ -216,6 +216,9 @@ func CreatePullRequest(
 		if err := repo.BranchSetConfig(ctx, opts.BranchName, "av-pushed-ref", fmt.Sprintf("refs/heads/%s", opts.BranchName)); err != nil {
 			return nil, err
 		}
+		if err := repo.BranchSetUpstream(ctx, opts.BranchName, remote); err != nil {
+			return nil, err
+		}
 	} else {
 		_, _ = fmt.Fprint(os.Stderr,
 			"  - skipping push to GitHub",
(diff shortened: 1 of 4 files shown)


internal/git/gittest/repo.go
```diff
@@ -105,6 +105,8 @@ func NewTempRepoWithGitHubServer(t *testing.T, serverURL string) *GitTestRepo {
 github:
     token: "dummy_valid_token"
     baseUrl: %q
+pullRequest:
+    branchNamePrefix: ""
```
willschnicke-silkline: my global av setting was conflicting, causing tests to fail. let me know if there is a more correct way to do this!

PR #679, opened 2026-03-20
fix: adopt --remote fails when branches aren't ordered parent-first

simsinght:
## Summary
- `av adopt --remote` fails with "ancestor branch is missing from av metadata" when adopting stacks
- Root cause: `initAdoption` built the branches slice by iterating `chosenTargets` from the tree selector, which has no guaranteed order. The remote PR walker returns PRs leaf-first, but `chosenTargets` doesn't preserve that order. When a child branch was adopted before its parent, `ValidateNoCycle` failed because the parent wasn't in metadata yet.
- Fix: iterate `prs` in reverse (parent-first) filtered to chosen targets, instead of iterating `chosenTargets` directly. Only the `--remote` codepath is affected; local adopt is unchanged.

## Test plan
- [x] Existing `TestAdopt` e2e tests pass
- [x] `pre-commit run --all-files` passes
- [x] Manually verified with a 7-branch stack (`av adopt --remote inbox/suspense`) that previously failed

### DIFF
--- a/cmd/av/adopt.go
+++ b/cmd/av/adopt.go
@@ -373,21 +373,21 @@ func (vm *remoteAdoptViewModel) initGitFetch(prs []actions.RemotePRInfo, chosenT
 }
 
 func (vm *remoteAdoptViewModel) initAdoption(prs []actions.RemotePRInfo, chosenTargets []plumbing.ReferenceName) tea.Cmd {
+	chosenSet := make(map[string]bool)
+	for _, target := range chosenTargets {
+		chosenSet[target.Short()] = true
+	}
 	prMap := make(map[string]actions.RemotePRInfo)
 	for _, prInfo := range prs {
 		prMap[prInfo.Name] = prInfo
 	}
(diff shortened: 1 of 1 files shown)


cmd/av/adopt.go
```diff
@@ -373,21 +372,22 @@ func (vm *remoteAdoptViewModel) initGitFetch(prs []actions.RemotePRInfo, chosenT
...
 		prMap[prInfo.Name] = prInfo
 	}
 	var branches []actions.AdoptingBranch
-	for _, target := range chosenTargets {
-		idx := slices.IndexFunc(prs, func(prInfo actions.RemotePRInfo) bool {
-			return prInfo.Name == target.Short()
-		})
-		if idx == -1 {
-			return uiutils.ErrCmd(fmt.Errorf("internal error: failed to find PR info for branch %s", target.Short()))
+	for i := len(prs) - 1; i >= 0; i-- {
```
tulioz: lets use `slices.Backward(prs)` here instead of the index loop

PR #709, opened 2026-04-07
feat(reorder): add squash/fixup commands and autosquash for fixup! commits

### LINKED ISSUE #706: Feature request: Allow squashing via `av reorder`
I would like to be able to squash commits during an `av reorder`, similar to how `git rebase` allows squashing commits.

Ideally, it should also process `fixup!` commits (generated by `git commit --fixup`) - these should be squashed against the commit that they fixup against.

amartani:
Add squash (s) and fixup (f) commands to av reorder, matching git rebase -i behavior. squash folds the cherry-picked commit into the previous one and opens the editor with combined messages; fixup does the same but silently keeps the previous commit's message.

Commits created with git commit --fixup (fixup! prefix) or git commit --squash (squash! prefix) are automatically repositioned after their target commit and pre-assigned the appropriate action when the plan is generated, matching git's --autosquash behavior.

Also handles conflict continuation correctly: when av reorder --continue is run after a squash/fixup conflict, the folding step is performed before advancing past the interrupted command.

Fixes #706


internal/reorder/plan.go
```diff
@@ -82,5 +83,97 @@ func CreatePlan(
...
+			fixPick.Mode = fixups[fixIdx].mode
+			result = append(result, fixPick)
+		}
+	}
+
+	// Append fixups whose target was not found anywhere in the stack,
+	// annotated with a warning so the user can see the problem in the editor.
+	for _, fixIdx := range unplaced {
+		fixPick := cmds[fixIdx].(PickCmd)
+		fixPick.Mode = fixups[fixIdx].mode
```
draftcode, line 170: Perhaps this is better to be PickModePick so that it won't accidentally amend the commit prior to that branch?

amartani: Ah good call; done in f2c30b9.

Also added 3e77b87 to use --no-verify when committing fixups.

draftcode (approved):

PR #713, opened 2026-04-09
fix nightly homebrew publish, pin goreleaser tag, disable fury for nightly

tulioz:
skip=publish was skipping all publishing including homebrew. use disable template on fury publisher instead, pin GORELEASER_CURRENT_TAG to avoid stale tag issues.


simsinght (approved):
arrghs

Issue #656, opened 2026-01-28
`av sync` needs to provide a better error message when branch can't be deleted due to worktrees


tulioz:
If you have a branch checked out on a worktree, and that branch has been merged, we attempt to delete the branch and fail with the error:
`error: cannot delete merged branch "more_ci_improvements": git branch: exit status 1`

```
❯ git switch more_ci_improvements
fatal: 'more_ci_improvements' is already used by worktree at '/worktrees/pr-feedback'
```

We should provide a better error message to the user to explain the cause and potentially give advice on how they can recover from this situation.

We should also avoid leaving the user in a detached head state when `av sync` fails to delete a branch

PR #719, opened 2026-04-17
fix(sync): include git stderr in branch-delete fatal error message

mvanhorn:
## Summary

Partially addresses #656. When `av sync --prune` (or any code path hitting `runDelete` in `internal/git/gitui/prune.go`) fails to delete a merged branch for a reason *other than* worktree ownership, the error message now includes the git stderr instead of the opaque `exit status 1`.

## Why this matters

Issue #656 reports the user-facing error:

```
error: cannot delete merged branch "more_ci_improvements": git branch: exit status 1
```

The "cannot delete merged branch" wrapper text traces to line 235 of `prune.go`, where the fatal-error path formatted the underlying error with `%v`. Since the error is an `*exec.ExitError`, `%v` resolves to `exit status 1` — stripping off the stderr git wrote to explain the failure. PR #660 already handled the specific "used by worktree" phrasing by catching that stderr and producing a recovery-oriented message, but any other failure reason still fell into this opaque path.

## What changed

`internal/git/gitui/prune.go` — inside `runDelete()`, the `else` branch (non-worktree fatal error) now pulls `exiterr.Stderr` off the `*exec.ExitError` (same `errutils.As` helper the worktree check above already uses) and passes it to the error message:

```go
errMsg := err.Error()
if exiterr, ok := errutils.As[*exec.ExitError](err); ok {
    if stderr := strings.TrimSpace(string(exiterr.Stderr)); stderr != "" {
        errMsg = stderr
    }
}
deletionErr = errors.Errorf("cannot delete merged branch %q: %s", branch.branch.Short(), errMsg)
```

So the user now sees whatever git actually said rather than the wrapped shell status.

## Scope notes

- The existing worktree-error handling (added in #660) is untouched; its friendlier recovery message still applies to "used by worktree" stderr.
- The detached-HEAD concern from #656's second paragraph is already handled by the unconditional `CheckoutInitialState()` call after the delete loop; no change needed there.
- Marking this as `Part of #656` rather than `Closes` because the issue's language covers worktree-specific UX that was delivered in #660, and this PR fills the remaining surface for non-worktree failures.

## Testing

```
go vet ./...           # pass
go build ./...         # pass
go test ./internal/git/... ./cmd/av/...   # pass
```

Part of #656

This contribution was developed with AI assistance (Codex).

### DIFF
--- a/internal/git/gitui/prune.go
+++ b/internal/git/gitui/prune.go
@@ -234,8 +234,15 @@ func (vm *PruneBranchModel) runDelete() tea.Msg {
 				worktreeBranches = append(worktreeBranches, branch.branch.Short())
 				continue
 			} else {
-				// Other errors are fatal
-				deletionErr = errors.Errorf("cannot delete merged branch %q: %v", branch.branch.Short(), err)
+				// Other errors are fatal. Surface the git stderr so users see the
+				// actual cause rather than just "exit status 1".
+				errMsg := err.Error()
+				if exiterr, ok := errutils.As[*exec.ExitError](err); ok {
+					if stderr := strings.TrimSpace(string(exiterr.Stderr)); stderr != "" {
+						errMsg = stderr
+					}
+				}
+				deletionErr = errors.Errorf("cannot delete merged branch %q: %s", branch.branch.Short(), errMsg)
 				break
 			}
 		}
--- a/internal/git/gitui/prune.go
+++ b/internal/git/gitui/prune.go
@@ -227,24 +227,23 @@ func (vm *PruneBranchModel) runDelete() tea.Msg {
 	for i := len(vm.deleteCandidates) - 1; i >= 0; i-- {
 		branch := vm.deleteCandidates[i]
 		if err := vm.repo.BranchDelete(context.Background(), branch.branch.Short()); err != nil {
-			// Check if the error is due to the branch being checked out in a worktree
-			if exiterr, ok := errutils.As[*exec.ExitError](err); ok &&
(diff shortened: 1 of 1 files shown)


internal/git/gitui/prune.go
```diff
@@ -234,8 +234,15 @@ func (vm *PruneBranchModel) runDelete() tea.Msg {
...
-				deletionErr = errors.Errorf("cannot delete merged branch %q: %v", branch.branch.Short(), err)
+				// Other errors are fatal. Surface the git stderr so users see the
+				// actual cause rather than just "exit status 1".
+				errMsg := err.Error()
+				if exiterr, ok := errutils.As[*exec.ExitError](err); ok {
+					if stderr := strings.TrimSpace(string(exiterr.Stderr)); stderr != "" {
+						errMsg = stderr
+					}
+				}
+				deletionErr = errors.Errorf("cannot delete merged branch %q: %s", branch.branch.Short(), errMsg)
```
tulioz: we're calling `errutils.As[*exec.ExitError](err)` twice on the same err here, once above for the worktree check and once here for stderr. can we just extract stderr once at the top of the error block, then branch on strings.Contains(stderr, "used by worktree") for the worktree case and fall through to the fatal path otherwise? also lets us drop the else after continue.

mvanhorn:
@tulioz Good call — pushed 47f68e0 to apply the refactor. stderr is now pulled out once at the top of the error block, the worktree branch uses `strings.Contains(stderr, "used by worktree")`, and the else after `continue` is gone.

tulioz (approved):

PR #736, opened 2026-04-29
fix(reorder): use full commit hashes to avoid short hash collisions

amartani:
During a reorder, build a short-to-full hash map in EditPlan (before showing the editor to the user) and resolve them back to full hashes when parsing, so git commands always run with full SHAs. User-introduced hashes not in the map are passed through unchanged. This avoids collisions that can happen on larger repos, while still maintaining user-friendly short hashes during editing.

Potential alternative: `git.ShortSha()` could use a longer prefix (12-digit is generally recommended for large repos), or it could run `git rev-parse --short $commithash` instead, to which git will always produce an unambiguous prefix for the repository. I think this mapping would still be appropriate to use, though.


internal/reorder/plan.go
```diff
@@ -67,25 +70,45 @@ func CreatePlan(
...
 
 		cmds = append(cmds, branchCmd)
 		for _, object := range commitObjects {
 			commit, err := git.ParseCommitContents(object.Contents)
 			if err != nil {
-				return nil, errors.WrapIff(err, "parsing commit %s", object.OID)
+				return nil, nil, errors.WrapIff(err, "parsing commit %s", object.OID)
 			}
+			shortHash := git.ShortSha(object.OID)
+			shortToFull[shortHash] = object.OID
```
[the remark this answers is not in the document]
amartani: 7-digit collision on a single stack is extremely unlikely.

internal/reorder/plan.go
```diff
@@ -67,25 +70,45 @@ func CreatePlan(
...
+	return autosquashCmds(cmds), shortToFull, nil
+}
+
+// ResolveHashes replaces short commit hashes in PickCmd entries with their
+// full hashes using the provided short-to-full map. Hashes not present in the
+// map (e.g., manually introduced by the user) are left as-is.
+func ResolveHashes(cmds []Cmd, shortToFull map[string]string) []Cmd {
+	resolved := make([]Cmd, len(cmds))
+	for i, cmd := range cmds {
+		if p, ok := cmd.(PickCmd); ok {
```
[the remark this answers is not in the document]
amartani: This is weird; it should be added only if actually needed in the future.

internal/reorder/plan.go
```diff
@@ -67,25 +70,45 @@ func CreatePlan(
...
 	// Reorder fixup!/squash! commits to sit immediately after their targets
 	// across the entire stack, matching git's --autosquash behavior.
-	return autosquashCmds(cmds), nil
+	return autosquashCmds(cmds), shortToFull, nil
+}
+
+// ResolveHashes replaces short commit hashes in PickCmd entries with their
+// full hashes using the provided short-to-full map. Hashes not present in the
+// map (e.g., manually introduced by the user) are left as-is.
+func ResolveHashes(cmds []Cmd, shortToFull map[string]string) []Cmd {
```
tulioz: could we just keep full hashes in the plan throughout and let `EditPlan` own the short/full mapping locally? render shorts for the editor, resolve back when parsing. then `CreatePlan` keeps a single return, `reorder.go` is unchanged, and `ResolveHashes` goes away. short hashes only matter for the editor display.

amartani: Done - changed PickCmd.String() to shorten the commit hash, and changed ParseCmd to also receive the shortToFull map to be able to rebuild the original Cmds. This seemed better than trying to do the actual shortening / restore in EditPlan itself since that would require basically re-parsing the strings in it. PTAL

internal/reorder/pick.go
```diff
@@ -193,7 +193,7 @@ func (p PickCmd) String() string {
 	}
 	sb.WriteString(mode)
 	sb.WriteString(" ")
-	sb.WriteString(p.Commit)
+	sb.WriteString(git.ShortSha(p.Commit))
```
tulioz: looks like the state round-trip breaks here. `MarshalJSON` writes `cmd.String()` (shorts), `UnmarshalJSON` parses with nil so they stay short, and `--continue` would cherry-pick 7-char hashes. the round-trip test passes because its inputs are 7 chars, but full hashes from `CreatePlan` wouldn't survive. could we keep `String()` as the full-hash form and do the short rendering inside `EditPlan` instead?

amartani: > could we keep String() as the full-hash form and do the short rendering inside EditPlan instead?

Code-wide, `EditPlan` gets a string (from `cmd.ToString()`), so to do the shortening there, we would need to essentially duplicate the parsing logic there to find where the commit hash is to be able to shorten it, which seems weird...

I added a new EditorString() method so that we can clearly differentiate when we are using the complete representation versus the shortened one. See if this makes sense. Also updated the round-trip tests to use the full commits. PTAL

tulioz (approved):

amartani:
@tulioz added a reformatting commit for the pre-commit issue

PR #737, opened 2026-04-29
fix(sequencer): clear orphaned state file after `git rebase --abort`

### LINKED ISSUE #734: `av restack`/`sync` reuses stale sequencer state after `git rebase --abort`, causing spurious conflicts
## Problem

When `av sync` or `av restack` hits a conflict, it writes sequencer state to `.git/av/stack-sync-v2.state.json` / `.git/av/stack-restack.state.json` so the operation can resume. If the user runs `git rebase --abort` directly (instead of `av restack --abort`), the state file is **not** cleared. Subsequent `av restack` / `av reparent` / `av sync` calls read the cached `BranchingPointCommitHash` from the state file and use it as the rebase old-base — even when the branch's real merge-base has moved and `av.db`'s `parent.head` has been updated. The result: av replays dozens of already-applied commits and produces spurious conflicts.

## Repro

1. Branch `X` is stacked on parent `P`. `av.db` records `parent.head = P_old`.
2. Manually run `git rebase --onto P X^^ X` — `X` is now cleanly on top of current `P`. `av.db` is unchanged.
3. `av sync` (or `av restack`) runs, captures `OriginalBranchSnapshots` with `BranchingPointCommitHash = P_old`, then hits a conflict mid-rebase.
4. User runs `git rebase --abort`. The state file persists.
5. Subsequent `av restack` / `av reparent` / `av sync` re-reads the state file and uses cached `P_old` as the rebase base, replaying the full `P_old..X` history onto `P_current`. Conflict storm.

Manually deleting `.git/av/stack-restack.state.json` + `.git/av/stack-sync-v2.state.json` resolves it.

## Root cause

`internal/sequencer/sequencer.go:192-206` picks the rebase base with **cached > fresh**:

```go
if snapshot.BranchingPointCommitHash.IsZero() {
    // compute fresh git merge-base
} else {
    // use the cached snapshot value
}
```

The snapshot is populated from `av.db`'s `parent.head` at sequencer construction (`sequencer.go:68-79`, `getBranchSnapshots` at `sequencer.go:81-95`) and persisted via `internal/git/state_file.go:26-40`. The state file is cleaned up only via the av abort path (`sequencer.go:135-144`, `internal/sequencer/sequencerui/ui.go:78-80`) — there's no detection of `git rebase --abort` performed outside av. Additionally, `av.db`'s `parent.head` is never re-read into an in-flight snapshot, so even out-of-band updates don't influence the rebase base.

## Proposed fixes

Any one of these would have prevented the stuck state:

1. **Detect a stale state file at startup.** Before honoring a pre-existing sequencer state, check for an in-progress `git rebase` (`.git/rebase-merge/` / `.git/rebase-apply/`, or `git status --porcelain=v2 -b`). If the state file claims we're mid-sync but no rebase is live, treat it as orphaned and clear it (with an info message). Smallest fix; maps 1:1 to the trigger.
2. **Prefer fresh merge-base over cached snapshot for unresumed branches.** In `sequencer.go:192-206`, if `av.db`'s `parent.head` differs from the snapshot's `BranchingPointCommitHash`, fall back to `git merge-base`. Also catches manual `av.db` edits.
3. **Clear state after a successful non-resume operation.** If a fresh invocation isn't a resume and the tree is clean, clear the state file before starting.

(1) is the most conservative and targets the exact trigger.

## Workaround

```bash
rm -f .git/av/stack-sync-v2.state.json .git/av/stack-restack.state.json
av restack
```

If `av.db`'s `parent.head` is also stale (e.g., from a manual `git rebase --onto` outside av), update it to the parent branch's current tip first.

simsinght:
## Summary

When `av sync`/`av restack` hits a rebase conflict, sequencer state is persisted to `.git/av/stack-{sync-v2,restack}.state.json` so the operation can resume. If the user runs `git rebase --abort` directly (instead of `av {sync,restack} --abort`), the rebase ends but the av state file is left behind. Subsequent `av {sync,restack}` invocations honor the orphaned state — re-using its cached `CurrentSyncRef` and `BranchingPointCommitHash` — and replay the same stale rebase, producing spurious conflicts.

## Fix

- Add `Repo.IsRebaseInProgress()` in `internal/git/git.go` — checks for `.git/rebase-merge/` or `.git/rebase-apply/` directories (both removed by `git rebase --abort`).
- In `cmd/av/restack.go` and `cmd/av/sync.go`, after loading a state file in `readState()`, verify a rebase is actually in progress. If not, treat the state as orphaned, clear the file, and return `nil` so the caller plans fresh from `av.db`.
- `cmd/av/reparent.go` is unchanged — it already always creates fresh state and doesn't read the file.

This corresponds to option (1) from the design sketch in #734 — the most conservative fix targeting the exact trigger.

## Reproduction

`e2e_tests/testdata/script/restack_external_rebase_abort.txtar` reproduces the bug:

1. Stack: `main → stack-1 → stack-2`.
2. Add a conflicting commit on `stack-1`.
3. `av restack` from `stack-2` → conflicts, persists state file.
4. `git rebase --abort` → REBASE_HEAD cleared, av state file orphaned.
5. `git checkout stack-1`.
6. `av restack --current` (no-op for `stack-1` as it has no children to restack).
   - **Without fix:** orphan is honored; `CurrentSyncRef=stack-2` is reused; same conflict reproduced. Test fails.
   - **With fix:** orphan detected and cleared; fresh `--current` plan runs and reports done. Test passes.

Verified locally: test fails on master, passes with the fix.

## Test plan

- [x] `go build ./...`
- [x] `go tool golangci-lint run` (0 issues)
- [x] `go test ./e2e_tests/` (all pass)
- [x] `go test ./internal/git/... ./internal/sequencer/... ./cmd/av/...` (all pass)
- [x] New regression test fails on master, passes on this branch

Closes #734

### DIFF
--- a/cmd/av/sync.go
+++ b/cmd/av/sync.go
@@ -320,6 +320,12 @@ func (vm *syncViewModel) readState() (*savedSyncState, error) {
 	} else if err != nil {
 		return nil, err
 	}
+	if !vm.repo.IsRebaseInProgress() {
+		if err := vm.repo.WriteStateFile(git.StateFileKindSyncV2, nil); err != nil {
+			return nil, err
+		}
+		return nil, nil
+	}
 	return &state, nil
 }
 
--- a/cmd/av/reorder.go
+++ b/cmd/av/reorder.go
@@ -59,6 +59,22 @@ squashed, dropped, or moved within the stack.
 			return err
 		}
 
+		// If state was loaded but no cherry-pick is in progress, the previous
+		// reorder was aborted externally (e.g., via `git cherry-pick --abort`).
+		// Treat the state as orphaned: clear it and proceed as if no state existed.
+		if continuation.State != nil && !repo.IsCherryPickInProgress() {
+			if err := repo.WriteStateFile(git.StateFileKindReorder, nil); err != nil {
+				return err
+			}
+			continuation = reorder.Continuation{}
+			if reorderFlags.Continue || reorderFlags.Abort {
+				fmt.Fprint(os.Stderr,
+					colors.Failure("ERROR: no reorder in progress\n"),
+				)
+				return actions.ErrExitSilently{ExitCode: 127}
+			}
+		}
+
 		var state *reorder.State
 		if reorderFlags.Abort {
 			if continuation.State == nil {
--- a/cmd/av/restack.go
+++ b/cmd/av/restack.go
@@ -163,6 +163,12 @@ func (vm *restackViewModel) readState() (*sequencerui.RestackState, error) {
 	} else if err != nil {
 		return nil, err
 	}
+	if !vm.repo.IsRebaseInProgress() {
+		if err := vm.repo.WriteStateFile(git.StateFileKindRestack, nil); err != nil {
+			return nil, err
+		}
+		return nil, nil
+	}
 	return &state, nil
 }
 
--- a/e2e_tests/testdata/script/restack_external_rebase_abort.txtar
+++ b/e2e_tests/testdata/script/restack_external_rebase_abort.txtar
@@ -1,0 +1,39 @@
+# Test that aborting a rebase via `git rebase --abort` (instead of `av restack --abort`)
+# does not leave behind a stale sequencer state file that breaks subsequent commands.
+#
+#     stack-1: main -> 1a
+#     stack-2:           \ -> 2a
+
+exec av branch stack-1
+commit-file my-file '1a\n' 'Commit 1a'
+exec av branch stack-2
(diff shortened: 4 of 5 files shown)


cmd/av/sync.go
```diff
@@ -320,6 +320,12 @@ func (vm *syncViewModel) readState() (*savedSyncState, error) {
 	} else if err != nil {
 		return nil, err
 	}
+	if !vm.repo.IsRebaseInProgress() {
```
jainankit, line 323: we should do some check in av reoder as well.

jainankit (approved):

PR #743, opened 2026-05-13
feat(sync): fast-forward branches behind remote

draftcode:
## Summary

When the remote tracking branch has commits on top of a local branch (i.e. local is strictly an ancestor of remote), `av sync` now fast-forwards the local branch to match remote before restacking. This lets users adopt commits pushed by collaborators (or by themselves from another machine) instead of clobbering them on the next push.

Diverged or local-ahead branches are left untouched — the check is strictly `merge-base --is-ancestor local remote` with local != remote.

## Implementation

- New `gitui.FastForwardBranchesModel` mirrors the existing `FastForwardTrunkModel` pattern.
  - Uses `git merge --ff-only` when the current branch is the one being fast-forwarded, `git update-ref` otherwise.
  - Iterates over the same target branches the sync flow already computes.
  - Prints `✓ Fast-forwarded <branch> to match remote` for each branch advanced; stays silent when there's nothing to do.
- Wired into the sync flow between `GitHubFetch` and the restack sequencer so the newly adopted commits flow through the rebase.

## Behavior

- Always on; no flag, since it's a no-op when FF is not trivially safe.
- Trunk fast-forward continues to be opt-in via `--ff-trunk` / `config.Av.Sync.FastForwardTrunk` and runs at the end of sync as before.

### DIFF
--- a/internal/git/gitui/ff_branches.go
+++ b/internal/git/gitui/ff_branches.go
@@ -1,0 +1,113 @@
+package gitui
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/aviator-co/av/internal/git"
+	"github.com/aviator-co/av/internal/utils/colors"
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/go-git/go-git/v5/plumbing"
+)
+
+// FastForwardBranchesModel is a Bubbletea model that fast-forwards each local
+// branch in targetBranches whose remote tracking branch is strictly ahead of
+// the local branch. This lets users adopt commits that someone else pushed on
+// top of their branch.
+type FastForwardBranchesModel struct {
+	repo           *git.Repo
+	targetBranches []plumbing.ReferenceName
+	onDone         func() tea.Cmd
+
+	ffBranches []string
+	done       bool
+}
+
+type ffBranchesDone struct{}
+
+func NewFastForwardBranchesModel(
+	repo *git.Repo,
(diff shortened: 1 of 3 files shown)


internal/git/gitui/ff_branches.go
```diff
@@ -0,0 +1,113 @@
...
+		remoteRef := fmt.Sprintf("refs/remotes/%s/%s", remote, name)
+		remoteExists, err := m.repo.DoesRefExist(ctx, remoteRef)
+		if err != nil || !remoteExists {
+			continue
+		}
+
+		// Skip if local is not an ancestor of remote (diverged or local is ahead).
+		isBehind, err := m.repo.IsAncestor(ctx, localRef, remoteRef)
+		if err != nil || !isBehind {
+			continue
```
jainankit, line 66: might make sense to show a warning to the user? I guess this would also happen in case of an amend? Is there a way to know if we are truly overwriting a remote commit?

internal/git/gitui/ff_branches.go
```diff
@@ -0,0 +1,113 @@
...
+func (m *FastForwardBranchesModel) run() tea.Msg {
+	ctx := context.Background()
+	remote := m.repo.GetRemoteName()
+
+	currentBranch, err := m.repo.CurrentBranchName()
+	if err != nil {
+		currentBranch = ""
+	}
+
+	for _, br := range m.targetBranches {
```
jainankit, line 54: No tests?

PR #746, opened 2026-05-15
Handle worktreeConfig with go-git v5

### LINKED ISSUE #745: error: failed to open git repo: failed to open git repo: core.repositoryformatversion does not support extension: worktreeconfig
I get this error when trying to use av cli on repos touched by the Codex app.

It seems this was fixed in go-git v6: https://github.com/go-git/go-git/pull/1877

matixlol:
## Summary
- keep the existing go-git v5.18.0 dependency; no forked module and no vendored copy
- ask `git rev-parse` for both the per-worktree git dir and common git dir, then open go-git with its public filesystem storage APIs
- wrap only config loading so v5 ignores the known `extensions.worktreeConfig` marker instead of rejecting the repository
- add regressions for normal and linked worktrees with `extensions.worktreeConfig=true`

Fixes #745.

## Validation
- `go test ./internal/git/... -count=1`
- `go build ./...`
- `go tool golangci-lint run`

### DIFF
--- a/cmd/av/helpers.go
+++ b/cmd/av/helpers.go
@@ -25,6 +25,7 @@ func getRepo(ctx context.Context) (*git.Repo, error) {
 			"rev-parse",
 			"--path-format=absolute",
 			"--show-toplevel",
+			"--git-dir",
 			"--git-common-dir",
 		)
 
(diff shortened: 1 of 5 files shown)


tulioz:
we've gone ahead and merged go-git v6

PR #749, opened 2026-05-16
fix(adopt): handle CTRL+C cleanly during av adopt (#692)

### LINKED ISSUE #692: Bug: When you CTRL+C when running av adopt, nothing happens
Only way to stop it appears to be to get the process ID and run `kill -9` on it.

mvanhorn:
## Summary
`av adopt` now exits cleanly on SIGINT (CTRL+C). Threads `context.Context` through the adopt path, switches git subprocess invocations to `exec.CommandContext`, and adds a shared `signal.NotifyContext` install at the Cobra entry.

## Why this matters
From #692:

> Only way to stop it appears to be to get the process ID and run `kill -9` on it.

Other av commands (`av sync`) honor SIGINT correctly, so the fix mirrors that pattern.

## Changes
- `cmd/av/main.go`: install `signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)` once at the Cobra root
- `cmd/av/adopt.go`: propagate the context into the adopt action
- `internal/actions/adopt_branches.go`, `find_adoptable_local_branches.go`, `get_remote_stacked_pr.go`, `git_fetch.go`: honor `ctx.Done()` in branch-walk loops
- `internal/treedetector/detector.go`, `internal/utils/uiutils/tea.go`: thread context through
- `e2e_tests/adopt_test.go`: new test spawning `av adopt`, sending SIGINT, asserting bounded exit time

Partial-state cleanup is correct: no half-adopted branches left in inconsistent state on cancellation.

Fixes #692

### DIFF
--- a/cmd/av/adopt.go
+++ b/cmd/av/adopt.go
@@ -70,14 +70,16 @@ The command will adopt the stack of pull requests starting from the specified br
 			if err != nil {
 				return err
 			}
-			return uiutils.RunBubbleTea(&remoteAdoptViewModel{
+			return uiutils.RunBubbleTeaWithContext(ctx, &remoteAdoptViewModel{
+				ctx:        func() context.Context { return ctx },
 				repo:       repo,
 				db:         db,
 				ghClient:   client,
 				branchName: adoptFlags.RemoteBranchName,
 			})
 		}
-		return uiutils.RunBubbleTea(&adoptViewModel{
+		return uiutils.RunBubbleTeaWithContext(ctx, &adoptViewModel{
+			ctx:               func() context.Context { return ctx },
 			repo:              repo,
 			db:                db,
 			currentHEADBranch: plumbing.NewBranchReferenceName(currentBranch),
@@ -143,6 +145,7 @@ func adoptForceAdoption(
 }
 
 type adoptViewModel struct {
+	ctx               func() context.Context
 	repo              *git.Repo
 	db                meta.DB
 	currentHEADBranch plumbing.ReferenceName
@@ -162,6 +165,7 @@ func (vm *adoptViewModel) Init() tea.Cmd {
 func (vm *adoptViewModel) initModel() tea.Cmd {
 	return vm.AddView(
 		actions.NewFindAdoptableLocalBranchesModel(
+			vm.ctx(),
 			vm.repo,
 			vm.db,
 			vm.initTreeSelector,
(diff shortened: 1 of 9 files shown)


mvanhorn:
Rebased onto master (master picked up the testscript e2e migration in #695) and pushed eaee61d to address the HIGH-priority data race.

State mutations on `GetRemoteStackedPRModel` (`m.failed`, `m.prs`, `m.done`) are now routed through `Update()` via two new message types (`remoteStackedPRFailedMsg`, `remoteStackedPRDoneMsg`). The `Init()` goroutine collects PRs into a local slice and only emits messages; all field writes happen on the Bubble Tea main thread, matching the framework contract. Build, `go test ./internal/actions/...`, and `golangci-lint` all clean.

The `e2e_tests/adopt_test.go` `TestAdopt_ExitsOnSIGINT` addition was dropped during rebase because master migrated to the testscript framework. I'll re-author it as a `.txtar` in a follow-up if you want explicit e2e SIGINT coverage; happy to fold that in here too if you'd prefer.

Not tackling the two medium suggestions (storing `context.Context` directly in `adopt_branches.go`) in this push so the data race fix lands clean. Will follow up.

mvanhorn:
Follow-up on the two medium gemini items.

**SIGINT e2e coverage**: pushed ba7620b adding `TestAdoptExitsOnSIGINT` in `e2e_tests/sigint_test.go`. It re-uses the merge-base sleep trick the original test had: spawn `av adopt --dry-run` with PATH-injected fake git that sleeps on `merge-base`, wait for the marker confirming the sleep started, send `os.Interrupt`, and assert `av` returns within 5s. Locks in the SIGINT path through `signal.NotifyContext` -> `RunBubbleTeaWithContext` -> the per-model ctx checks. Test passes in ~0.7s. The file deliberately sits outside the testscript harness - testscript v1.10.0 in go.mod has no public API for sending an arbitrary signal to a specific named background process (it only sends `os.Interrupt` to every backgrounded command at end-of-test via `waitBackground`). A header comment in the file documents the conditions under which this can migrate into `testdata/script` as a `.txtar`.

**context.Context direct storage on `AdoptBranchesModel` / `GetRemoteStackedPRModel`**: tried the suggested refactor, both models then trip the `containedctx` golangci-lint rule that the repo already enforces. Reverted - keeping the `func() context.Context` closure indirection consistent with the rest of the codebase. Happy to revisit if the lint rule changes.

PR #751, opened 2026-05-29
fix(sync): use per-worktree git dir for rebase detection and state files

jainankit:
## Summary

\`getRepo()\` (\`cmd/av/helpers.go\`) currently reads \`git rev-parse --git-common-dir\` and stores it as the one and only \`gitDir\` on the \`Repo\` struct. That's correct for \`av.db\` (which is shared across worktrees), but wrong for anything inherently per-worktree: \`rebase-merge/\`, \`rebase-apply/\`, \`CHERRY_PICK_HEAD\`, and av's own sync/restack/reorder state files. In a linked worktree these live under \`.git/worktrees/<name>/\`, not the common dir, so:

- \`IsRebaseInProgress()\` returns false even when \`git status\` correctly reports a conflict-mid-rebase.
- \`readState()\` (\`cmd/av/sync.go:315\`) sees the state file exists but \`IsRebaseInProgress()\` is false, **deletes the state file**, and reports \"no restack in progress\" on every subsequent \`av sync --continue\`.

End user impact: starting a sync in a worktree, hitting a conflict, and being unable to continue — the state av wrote a moment earlier is wiped on first \`--continue\`.

## What changed

- \`Repo\` gains a \`worktreeGitDir\` field; \`WorktreeGitDir()\` and \`WorktreeAvDir()\` getters added.
- \`OpenRepo(repoDir, gitDir, worktreeGitDir)\` — third arg is the per-worktree git dir.
- \`IsRebaseInProgress()\` and \`IsCherryPickInProgress()\` now look under \`WorktreeGitDir()\`.
- \`ReadStateFile\` / \`WriteStateFile\` now read/write under \`WorktreeAvDir()\`. \`ReadStateFile\` falls back to the legacy shared path so an in-flight sync started on the old av version survives the upgrade. \`WriteStateFile(kind, nil)\` clears both locations.
- \`getRepo()\` requests \`--git-dir\` in addition to \`--git-common-dir\` and passes both to \`OpenRepo\`.
- \`gittest.GitTestRepo.AsAvGitRepo()\` passes its single \`GitDir\` as both args (tests use regular checkouts).

\`av.db\` and \`AvTmpDir()\` continue to live under the common dir, as before.

## Test plan

- [x] \`go test ./...\` — full suite green.
- [x] New \`TestWorktreeStateIsolation\` — creates a linked worktree, asserts \`IsRebaseInProgress\` sees the worktree-local \`rebase-merge\` dir, asserts state files round-trip through the worktree av dir and never touch the common dir.
- [x] New \`TestReadStateFileLegacyFallback\` — state file at the legacy shared path is still readable after upgrade.


internal/git/state_file.go
```diff
@@ -15,26 +15,53 @@ const (
...
 func (r *Repo) ReadStateFile(kind StateFileKind, msg any) error {
-	bs, err := os.ReadFile(filepath.Join(r.AvDir(), string(kind)))
+	bs, err := os.ReadFile(r.stateFilePath(kind))
 	if err != nil {
+		// Fall back to the legacy shared-AvDir path so in-flight syncs from
+		// an older av version remain resumable after upgrade.
+		if os.IsNotExist(err) {
+			legacy := filepath.Join(r.AvDir(), string(kind))
+			if legacy != r.stateFilePath(kind) {
+				bs, err = os.ReadFile(legacy)
```
simsinght: this fallback only ever fires inside a linked worktree (in the main checkout `legacy == stateFilePath`, so the guard skips it). but `<common>/av` is also where the *main* worktree writes its state now — so this isn't just reading a pre-upgrade leftover, it reads the main worktree's live state as if it were ours.

chained with the orphan check that's the problem: fresh `av sync` in a worktree → no state of its own → falls back and loads main's → `IsRebaseInProgress` on *this* worktree is false → we call it orphaned → `WriteStateFile(nil)` below removes the legacy path too → main worktree's paused sync is gone.

repro'd it locally: leave a paused sync in the main worktree, run `av sync` in a linked one, main's state file is deleted.

migration intent makes sense, but i think we need to scope it so the fallback can't resolve to another worktree's active state, and not clear legacy on every single clear.

simsinght (commented):
core change is right — splitting `worktreeGitDir` off the common dir and pointing the detection helpers + state files at the per-worktree one fixes the #737 regression in worktrees.

one gap though: this only swept the two helpers (`IsRebaseInProgress` / `IsCherryPickInProgress`). the same root cause — reading a per-worktree git file out of the shared `GitDir()` — still bites in three other readers that aren't in this diff:

- `reorder.go:87` + `:105` — stats `GitDir()/CHERRY_PICK_HEAD` for `--abort`/`--continue`, so in a worktree both silently skip the cherry-pick. also a little inconsistent now: `:66` uses the new `IsCherryPickInProgress()` while these two right below still do the old stat.
- `sequencer.go:137` — stats `GitDir()/REBASE_HEAD` before `git rebase --abort`. nil in a worktree → `av sync --abort` skips the abort and leaves you mid-rebase.
- `internal.go:10` (`readGitFile`, only used for `CHERRY_PICK_HEAD` via `cherrypick.go`) — reads the common dir, so a cherry-pick conflict in a worktree returns "expected CHERRY_PICK_HEAD to exist" instead of the real conflict error.

none of these are introduced here — all pre-existing — so fine to punt to a follow-up. but they undercut the "worktrees work now" story and they're all the same one-line `GitDir()` → `WorktreeGitDir()` swap. confirmed each with a real worktree locally.

left a separate inline on `state_file.go` for the one thing that *is* new in this PR.

internal/git/state_file.go
jainankit: updated in this PR itself.

simsinght (approved):

PR #760, opened 2026-06-18
make repo close nil-safe

tulioz:
follow-up to #759, defensive nil guard on Repo.Close per review

### DIFF
--- a/cmd/av/main.go
+++ b/cmd/av/main.go
@@ -136,11 +136,7 @@ func run() int {
 	// runtime and various packages (e.g., package init functions).
 	startTime := time.Now()
 	colors.SetupBackgroundColorTypeFromEnv()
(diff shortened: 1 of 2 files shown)


cmd/av/main.go
```diff
@@ -136,11 +136,7 @@ func run() int {
 	// runtime and various packages (e.g., package init functions).
 	startTime := time.Now()
 	colors.SetupBackgroundColorTypeFromEnv()
-	defer func() {
-		if cachedRepo != nil {
```
jainankit, line 140: we don't need cachedRepo null check?

tulioz: yeah, `Close()` checks `r == nil` itself now so we don't need it here

jainankit (approved):

PR #766, opened 2026-07-02
speed up av sync on repos with many remote refs

tulioz:
on large github repos the unfiltered ls-remote (365k+ refs on one customer repo) and the 10k-commit go-git trunk walk make even no-op syncs take ~30s+

### DIFF
--- a/internal/git/gitui/prune.go
+++ b/internal/git/gitui/prune.go
@@ -19,6 +19,7 @@ import (
 	"github.com/erikgeiser/promptkit/selection"
 	"github.com/go-git/go-git/v6/config"
 	"github.com/go-git/go-git/v6/plumbing"
+	"github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


internal/git/gitui/prune.go
```diff
@@ -366,11 +364,17 @@ func (vm *PruneBranchModel) calculateMergedBranches() tea.Msg {
 			)
 			continue
 		}
-		remoteHash, ok := remoteBranches[fmt.Sprintf("refs/pull/%d/head", avbr.PullRequest.Number)]
-		if !ok {
+		remoteHash, err := vm.repo.FetchRemoteRef(
```
jainankit: this fetch is called in a for-loop, you could batch it.

jainankit (approved):
1 non-blocking suggestion

PR #769, opened 2026-07-07
fix(git): parse renamed file paths containing spaces in git status

devteamaegis:
## What's broken
`Repo.Status()` records the wrong path for renamed/copied files whose new path contains a space. For a real `git status --porcelain=v2` line like `2 R. N... 100644 100644 100644 <h> <h> R100 new file.txt<TAB>old.txt`, the parser stores `file.txt` instead of `new file.txt`.

## Why it happens
`patternFile2` used `... [0-9a-f]+ [0-9a-f]+ .+ (.+)\t.+`. The leading `.+` (meant to match the rename score field, e.g. `R100`) is greedy and `.` matches spaces, so the path was split on its last space and only the final segment was captured.

## Fix
The score field never contains a space, so match it with `\S+`. The path group then captures everything up to the tab, including embedded spaces.

## Test
Added `internal/git/status_test.go` with a rename-with-space case (fails before, passes after) and a no-space case. Full `internal/git/...` suite passes.


tulioz:
I do not see any associated bug. If you are having issues with the cli, please open an issue and we can discuss there.

